PR #11, opened 2026-04-29
0.14.2 wasm batch: no-alloc fast path + Request.query + edge demo polish

jasisz:
## Summary

Adds a shared per-fn allocation analysis (`src/ir/alloc_info.rs`) parametrised by a backend-specific `AllocPolicy` trait, then wires it into the WASM emitter and the VM bytecode compiler so functions that provably never produce a heap object skip GC framing entirely.

Closes the bulk of the pure-float perf gap with peer typed-wasm + Cranelift toolchains. The Mandelbrot bench (160×96 grid × 100 reps × 80 iter, mutual-TCO `mandelStep ↔ mandelIter` with cardioid + period-2 bulb pre-tests, identical algorithm and result `342500`) runs **~4× faster on WASM** and **~22% faster on VM**.

## Per-backend changes

**WASM (`emit_plain_user_function` + `emit_mutual_tco_trampoline` + `expr/emit.rs` tail-call site):**
- Skip the prologue `global.get $heap_ptr; local.set $mark` save and the epilogue `$rt_truncate` call when the body is no-alloc.
- Skip `gc_watermark` allocation, per-iter heap_ptr save, and the watermark-> 16384 compaction check inside trampolines whose every member is no-alloc.
- Collapse the trampoline slot table to a single shared row when all members share an identical wasm-typed signature (`mandelStep ↔ mandelIter`-style groups).
- Under (uniform layout && no-alloc), the per-tail-call args reshape skips its `eval → tmp → target` double-copy and writes args directly into target slots in reverse stack order.

**VM (`compiler/mod.rs` + `execute/dispatch.rs` + `types.rs`):**
- `compute_alloc_info` runs after the existing `classify_thin_functions` pass and tags qualifying chunks with `no_alloc = true` (and `thin = true` for the matching fast-return contract).
- `TAIL_CALL_KNOWN` and `TAIL_CALL_SELF` skip `finalize_frame_locals_for_tail_call` when the target/current chunk is no-alloc — replaced by a `copy_within` slot relocation since the finalizer would always fall through to its no-op branch.
- `CALL_KNOWN` parks dummy zero marks for no-alloc targets (skips three `young_len()` / `yard_len()` / `handoff_len()` reads).
- `RETURN` short-circuits the standard `can_fast_return` path and `flatten_deep_list` precaution for no-alloc frames — they produced no survivors to evacuate.

## Bench (in-process wasmtime / `aver run`, identical algorithm, identical result)

| backend | before | after | speedup |
| --- | --- | --- | --- |
| WASM (wasmtime real) | ~590 ms | ~150 ms | **~4×** |
| WASM (net compute, –110 ms baseline) | ~480 ms | ~40 ms | **~12×** |
| VM (`aver run` real) | ~2710 ms | ~2100 ms | **~22%** |
| Aver vs Vera ratio (apple-to-apple) | 4.2× slower | **1.07× slower** | gap liquidated |

`comparison_bench` summary (in-process VM/WASM only — codegen/self-hosted dominated by subprocess spawn variance):
- VM: -3% to -23% across 9 benches
- WASM: -1% to -7% on 8 benches; **+8% regression on `pattern match 30k`** (statistically significant, root cause not yet identified — known follow-up for 0.15)

## Tagged-i64 design preserved

Backends keep their existing Value representations (NaN-boxing on VM, tagged i64 on WASM). The shared analysis just lets each backend skip framing it would otherwise emit. Generic `List<T>` / `Map<K, V>` / `Vector<T>` runtime is unchanged.

## Stress tests

- 1057/1057 lib tests green
- 18/18 cross_backend_stress (VM ↔ WASM ↔ self-host parity)
- 30 non-game examples (calculator, fibonacci, quicksort, oracle, hostile-clock, …) on both VM and WASM: all `exit 0`, peak RSS unchanged (VM ~4–7 MB, WASM ~12.5 MB)
- 7 games (snake, wumpus, life, checkers, doom, rogue, tetris) verify + compile to WASM cleanly (18–54 KiB)
- `cargo clippy -p aver-lang --features wasm -- -D warnings` and `cargo fmt --check` clean
- `nan_value_bench` ortogonalne, nie dotknięte

## Test plan

- [x] `cargo test --features wasm` — full lib + integration suite
- [x] `cargo test --features wasm --test cross_backend_stress`
- [x] `cargo bench --bench comparison_bench --features wasm` — VM/WASM speedups confirmed, 1 known WASM regression noted
- [x] All 30 non-game examples on both backends
- [x] All 7 games verify + compile to WASM
- [x] Mandelbrot vs Vera apple-to-apple bench

## Known follow-ups (0.15)

- `pattern match 30k` WASM +8% regression — likely instruction-scheduling interaction from skipping `area`'s prologue/epilogue ops; needs `perf record` to root-cause
- Type-aware `AllocPolicy` (e.g. `Map.get<K, Int>` no-alloc vs `Map.get<K, String>` alloc) — current `builtin_allocates(name)` lacks `&Type` context, so `VmAllocPolicy` whitelist stays conservative
- Inline 2-member mutual-TCO into a single fused fn (eliminate trampoline + wrappers entirely) for full Vera-parity compute
- Prefix-shared trampoline signatures for non-identical groups


jasisz:
## Scope extension — two follow-up commits on this branch

While dogfooding the no-alloc work via `tools/edge` (the `/fractal` Cloudflare Worker demo) two adjacent items shook out and landed here rather than on a second PR:

### 1. `wasm: HttpRequest.query field + friendly error for HTTP under non-fetch` (ec09ef1)

- **`req.query: String` on `HttpRequest`.** Plumbed through `builtin_records.rs`, `abi.rs` (`Request.query` → `request_query`), the fetch-bridge field-emit map, the cloudflare `worker.js` template (`new URL(req.url).search.slice(1)`), the type-checker's net-request field list, and the unconditional fetch-adapter pre-import block in `emitter.rs`. Closes the asymmetry where `req.body` worked but `req.query` was silently dropped at the host boundary.

- **Friendly compile-time error for HTTP types under non-fetch adapters.** Previously `aver compile … --target wasm` (no bridge) on a program with `fn handler(req: HttpRequest) -> HttpResponse` produced invalid bytecode (`type mismatch: expected i64, found i32` deep in the record-shape codegen) — now the emitter rejects it before emit with a message naming the offending fn and pointing at `--preset cloudflare` / `--bridge fetch`.

### 2. `edge demo: 16-colour half-block fractal with query-driven pan/zoom` (a90fe56)

Reworks `/fractal` from a fixed-pair monochrome Braille render into a single-panel interactive Mandelbrot:

- Half-block (▀) cells with per-pixel colour. Each cell encodes top/bot palette indices in a 2-hex `c=XX` attribute (`<i c=07>▀</i>` = 15 B/cell vs the old 24 B). 32 CSS attribute selectors `i[c^=H]` / `i[c$=H]` cover all 256 combinations from one shared 16-step palette, ~38% smaller HTML.
- Single square-on-screen panel — CSS sizes the `<pre>` by `min(viewport-w, viewport-h)` so it always fits without scroll.
- Interactive nav: `cx` / `cy` / `w` come from `req.query` (the new field above). URL is the source of truth, every click fires a fresh handler invocation. Pan = ±½ viewport, zoom = ÷/×2, reset = fullView. Six landmark presets (seahorse, mini, spiral, elephant, tendrils, eastern bulb).
- Pure-Aver query parser (`String.split` + `Vector.fromList` linear scan, `Float.fromString` with default fallback).
- Grid stays 200×120 across all zoom depths; iter cap auto-scales (100 / 150 / 250) so deep zooms stay sharp at modest extra compute.
- `cache-control: no-store` — every navigation is a real render.

### Production numbers (CF Workers, edge.averlang.dev)

| view | grid | iter | warm TTFB |
| --- | --- | --- | --- |
| fullView | 200×120 | 100 | ~35-40 ms |
| mid (w=0.1) | 200×120 | 150 | ~37 ms |
| deep (w=0.012) | 200×120 | 250 | ~40-50 ms |

(Keep-alive measurement; first hit ~70 ms with TLS handshake.)

### Vera comparison update

Same Mandelbrot pure-compute bench (200×240 px × 20 renders, deep zoom + cardioid/bulb pre-tests, wasmtime):

| view | iter | Aver | Vera | gap |
| deep zoom | 220 | 25 ms | 22 ms | 1.11× |
| fullView | 100 | 3 ms | 2 ms | 1.5× |

The "Aver vs Vera ratio" row in the bench table above held — the float-tight-loop gap is now ~10-15%, not ~3×.

### Stress tests after these commits

- `cargo test --lib --release --features wasm` → 563/563 green
- `aver audit tools/edge --module-root tools/edge` → 3 files | 0 errors | 0 verify failures | 0 format
- Friendly-error path verified end-to-end (6 input combinations: bare `--target wasm`, `--bridge wasip1`, `--bridge fetch`, `--preset cloudflare`, non-HTTP programs, `tools/edge/app.av`)
- `req.query` end-to-end: nav anchors on `https://edge.averlang.dev/fractal?w=0.5` correctly reflect parsed query (zoom-in halves to `w=0.25`, pan preserves cx/cy, etc.)

PR #14, opened 2026-05-02
0.16 "Anneal" — Console=String, native wasm-gc, soundier internals

jasisz:
## Summary

Heat-treat the internals — every backend now agrees what `Console.print` is, what's a type, and that engine GC beats hand-rolling.

## Breaking change

**`Console.print` / `Console.error` / `Console.warn` / `Terminal.print` / `Terminal.setColor` now take `String`.** The previous polymorphic `Printable` signature is gone. Stringification is the caller's job — interpolation `"{x}"` for primitives, per-type render helpers for compound shapes (`fn show(r: Result<T, E>) -> String`).

```aver
# Before
Console.print(divide(10, 2))

# After
Console.print(showResult(divide(10, 2)))
```

Programs that called `Console.print` with non-String args fail typecheck with `Argument 1 of 'Console.print': expected String, got <Type>` and the message points at the call site to migrate. Every backend (VM, legacy wasm, wasm-gc, Rust, self-host) loses its per-type format dispatch; replay/record stores plain string bytes per call; the effect ABI becomes trivial across hosts.

Bundled migrations cover the typical patterns — see `examples/core/calculator.av` (Result-render), `examples/core/lists.av` (List-render via direct recursion), `examples/services/disk_demo.av` (count-not-dump), and the playground sources.

### `--target wasm-gc` is the recommended WASM target

Native WebAssembly GC + tail-call output. No NaN-boxing, no boundary GC framing, no inline runtime. Per-instantiation monomorphisation for `Vector<T>` / `List<T>` / `Map<K, V>` / `Option<T>` / `Result<T, E>` / `Tuple<A, B>` (incl. nested paren tuples like `(Int, (Int, Int))`).

Full Aver type parity: tuples, cross-collection nesting, sum-type and record map keys, all 12 `Terminal.*` effects, multi-binding variant patterns, `?` / `?!` / `Result<Unit, …>` / `Tuple<…, Unit, …>` shapes, generic constructors driven by bidirectional inference. Wins V8 microbenches on alloc / recursion / collection workloads (vector_ops 269×, map_build 5.67×, record_access 3.37×); `--optimize size` shrinks per-program binaries 19-32 %.

### `aver run --wasm-gc <file>`

Embedded wasmtime executor with engine GC + tail calls. Full effect surface wired against `aver_rt::*`:

| Namespace | Effects |
|---|---|
| `Args` | `len`, `get` |
| `Console` | `print`, `error`, `warn`, `readLine` |
| `Time` | `now`, `unixMs`, `sleep` |
| `Random` | `int`, `float` |
| `Float` | `sin`, `cos`, `atan2`, `pow` |
| `Terminal` | `enableRawMode`, `disableRawMode`, `clear`, `moveTo`, `print`, `setColor`, `resetColor`, `hideCursor`, `showCursor`, `flush`, `readKey`, `size` |
| `Disk` | `readText`, `writeText`, `appendText`, `exists`, `delete`, `deleteDir`, `listDir`, `makeDir` |
| `Env` | `get`, `set` |
| `Tcp` | `connect`, `writeLine`, `readLine`, `close`, `send`, `ping` |
| `Http` | `get`, `head`, `delete`, `post`, `put`, `patch` |

Multi-module entries auto-resolve their module root from the entry file's `depends [...]`, so `aver run --wasm-gc projects/payment_ops/main.av` works without `cd` or `--module-root`.

Of the 71 single-file or entry-point examples + project mains, **68 run identically to the VM** (every `examples/{core, data, games, formal, wasm, services, apps, modules}` and `projects/*` plus 13 of 14 playground samples). The remaining 3 are intentionally bad code that *show* compile errors (`examples/diagnostics/{lint_demo, test_errors}.av`, `tools/website/playground/sources/examples/effect-violation.av`).

### Browser playground migrated to `--target wasm-gc`

Engine GC replaces the `aver_runtime.wasm` sidecar; binaries are self-contained. Modern-browser baseline (Chrome 119+ / Firefox 120+ / Safari 18.2+).

### Soundier internals

- `Type::Unknown` / `Type::Any` removed from the static type enum. Generic builtin positions use named `Type::Var(K|V|T|E)` resolved per call site by `match_expected_type`; error recovery uses `Type::Invalid` (never compatible with concrete types).
- Bidirectional inference drives expected types into generic constructors and collection literals — `Map.empty`, `Option.None`, `Result.Ok/Err`, empty list literals all type without backend recovery shims, including in tail-call argument position.
- Resolver is arm-scoped: pattern bindings live per arm (`MatchArm.binding_slots`) instead of a function-flat name → slot map. Two arms can share a binding name with different field types — `recordedAt` reusing `at` across TaskEvent variants, `serializeTaskEvent` reusing `deadline` as `Option<String>` in one arm and `String` in another — and each arm allocates its own slot.
- Slot types lifted into IR as `FnResolution.local_slot_types` — every backend that needs typed locals consumes the same table instead of re-walking patterns. Sibling sumtypes sharing a bare variant name (e.g. `Query.ProviderSummary` vs `QueryOutput.ProviderSummary` in payment_ops) are disambiguated by `(parent, bare)` keys throughout resolver and wasm-gc registry.
- Multi-arm tuple-of-Constructors matches and chained-Attr value access (`Domain.Types.TaskStatus.Blocked`) flatten through one shared rewrite path used by both VM and wasm-gc.
- `Char.fromCode` accepts the full Unicode range (Doom Braille block, etc.).

### Known regression

`newtype record` reads measure 0.18× on V8 wasm-gc vs the legacy `--target wasm` path (~ns scale; wasm-gc is faster on every other workload). Tracking — not a release blocker.


aallan:
Downstream verified from vera-bench: the `Console.print` typed migration in this release doesn't break anything on our end. We landed forward-compat work in [aallan/vera-bench#65](https://github.com/aallan/vera-bench/pull/65) (released as [v0.0.11](https://github.com/aallan/vera-bench/releases/tag/v0.0.11) on 2026-05-04) before 0.16 was on crates.io. Pulled the released `aver-lang 0.16.0` overnight and re-ran the bench's Aver baselines:

```
Problems       60
check@1        100%
run_correct    100%
Tier 1–5       100% across all five tiers
```

Identical to the numbers we got on Aver 0.15.2 with the migration applied — interpolation form (`Console.print("{x}")`) holds up cleanly on the actually-released 0.16.0. No surprises between the `0.16-wasm-gc-probe` build the migration was developed against and the shipped release.

Thanks for the heads-up on the breaking change — the bench was 0.16-ready on day one of the release rather than scrambling after the fact, which is the right rhythm for a fast-moving language.

Issue #147, opened 2026-05-26
RFC: resolved identity layer phases B + E — staged plan (deferred from #138)


jasisz:
## Status

Issue #138 (resolved identity layer) closed phases A + C across PRs #141–#146:

- **Phase A** (proof codegen): `ProofIR.fn_contracts` / `refined_types` / `LawTheorem.fn_id` → opaque IDs; Dafny SCC sets keyed by FnId
- **Phase C** (codegen ctx-level): `CodegenContext.recursive_fns` / `mutual_tco_members` → `HashSet<FnId>`; `SymbolTable` always present in pipeline; `scc::*` boundary helpers

**Phases B / D / E deferred deliberately.** This issue captures the staged plan + triggers for picking them back up.

## Why deferring

Empirical evidence that the bug class doesn't exist in practice — the typechecker stops source code with cross-module same-bare-name collisions:

```aver
// A.av: type Shape { Circle(Float) }
// B.av: type Shape { Triangle(Float) }
// main.av:
fn aFn(s: A.Shape) -> Float = match s { A.Shape.Circle(r) -> r }
fn main() -> Float = aFn(A.Shape.Circle(1.0))
//                       ^ error: expected A.Shape, got Shape
```

`sig_aliases` resolves canonical name before strict compare. The wasm_gc string-keyed monomorph maps NEVER see a program with the collision — the typechecker throws an error earlier. Bug-fix value for B/D/E = 0 today.

Plus cost: ~400 sites of `Type::Named(String)` pattern matches in the typechecker, 141 callsites of `fn_sigs.get(...)`, 50+ cascading sites in signatures. 3-5 days of solid work with risk of regressions in the core front-end.

Governing memory items: "no premature optimization", "Quality over speed", "Don't add features beyond what the task requires".

## Why eventually doing (peer AI consultation)

> "This isn't premature optimization; it's closing the semantic model. In a real-world setting, stopping after A+C is reasonable. In a world of 'I have time and want to set up the compiler well for the coming years', I'd do B/E."

Aver is a foundation-first pre-1.0 language. B/E has no bug-fix trigger TODAY, but every future feature (LSP rename, inliner, monomorph, incremental compile) will force typed identity — then B/E becomes a natural prerequisite rather than a standalone refactor.

### Phase B — typechecker typed identity (incremental, ~2-3 days)

Not a brutal cascade `Type::Named(String) → Type::Named(TypeId)` straight away. Intermediate layer:

[...]

PR #148, opened 2026-05-26
types/checker: typed identity (Type::Named<TypeId>, fn_sigs by FnId) — #147 phase B

jasisz:
## Summary

Phase B of #147: the typechecker now carries opaque identity through the same `SymbolTable` layer the proof IR and backends already consume (#141–#146). Cross-module same-bare-name user types stay distinct by `TypeId` at the matcher boundary instead of relying on `sig_aliases` string→string canonicalisation.

Single atomic PR — design choices documented inline so the review can stay in one session.

## Design

Picked a single-variant approach rather than `Type::Unresolved` + `Type::Named`:

```rust
pub enum Type {
    …
    Named { id: Option<TypeId>, name: String },
}
```

- `id` is the load-bearing identity. Two `Named` with both `Some` compare by `id == id` (typed identity — cross-module `Shape` types stay distinct by construction).
- `name` stays **source-faithful** (what the user wrote: `"Shape"` or `"A.Shape"`), so downstream string-equality consumers — Rust codegen's boxing detection (`param_name == ret_name`), wasm-gc dispatch tables — keep seeing the pre-phase-B form. The matcher falls back to source-name comparison when either side's `id` is `None` (builtins like `HttpResponse` / `Header` / `Tcp.Connection`, in-flight stamps that didn't resolve).

The dual-variant `Unresolved` design was attempted earlier in the session but produced an explosion of pattern-match touch points across 38 files without a real soundness improvement over the `Option<TypeId>` field; ~127 mechanical sweeps were already painful, dual variants would have ~doubled the diff. Compiler eats complexity, not surface area.

### Storage split

`TypeChecker.fn_sigs: HashMap<String, FnSig>` is gone. In its place:

- `fn_sigs: HashMap<FnId, FnSig>` — **user-defined** functions only (one entry per `FnId` in the program symbol table).
- `extra_sigs: HashMap<String, FnSig>` — builtin namespace methods (`Int.add`, `Console.print`) and sum-type variant constructors keyed by canonical `"Module.Type.Variant"`. These don't have `FnId`s in `SymbolTable` because they're not part of the user-program identity space.

`find_fn_sig` / `insert_fn_sig` / `all_fn_sigs` chain the two halves so existing bare-name callsites (and the exported `TypeCheckResult.fn_sigs` string-keyed map) keep working.

### sig_aliases dropped

`sig_aliases: HashMap<String, String>` is gone. Replaced by typed bridges:

- `bare_fn_aliases: HashMap<String, FnId>`
- `bare_type_aliases: HashMap<String, TypeId>`

Populated from visibility-exposed aliases in `integrate_registry` (and from `build_signatures` for the current module's own fns). `resolve_fn_id` / `resolve_type_id` chain: literal-as-qualified → current-module-scoped → entry-scope → typed bare alias.

### Pipeline ordering

`SymbolTable::build` now runs **before** typecheck in `build_symbols_for_items` / `build_symbols_with_loaded`. The table is threaded into the checker via `TypeChecker::new_with_symbols`. The existing `pipeline::run` build of the table after typecheck stays — `result.symbol_table` is unchanged for backward compat, just built deterministically against the same inputs.

### Opaque ID types relocated

`FnId` / `TypeId` / `CtorId` / `ModuleId` moved from `ir::symbol_table` to `ir::identity` so `ast::Type` can carry a `TypeId` field without forming a cycle through `ir::symbol_table` (which depends on `ast`). Re-exports preserve every existing import path.

## Verified

- ✅ `cargo build --features wasm` — clean.
- ✅ `cargo test --features wasm` — **1608 passed, 0 failed, 7 ignored**.
- ✅ `cargo clippy --all-targets --features wasm -- -D warnings` — clean.
- ✅ `cargo fmt --check` — clean.
- ✅ Regression test `cross_module_same_named_types_do_not_merge` passes.
- ✅ `aver compile examples/core/order_total.av --target wasm-gc` produces **byte-identical output** to pre-change (md5 `61e45b325279e17e1b0d8dce3fde43f9`).
- ✅ `aver check` on `examples/games/snake.av` and `examples/games/rogue/main.av` succeeds.

## Test plan

- [ ] Skim the patch (~940 +/-494, 45 files; bulk is mechanical `Type::Named(n)` → `Type::Named { name: n, .. }` and `Type::Named("X".to_string())` → `Type::named("X")`)
- [ ] Spot-check `src/types/checker/mod.rs` — new `resolve_fn_id`/`resolve_type_id`/`canonical_type_name` helpers + storage split rationale
- [ ] Spot-check `src/types/checker/modules.rs` — `build_signatures` / `register_type_def_sigs` / `integrate_registry` route through `insert_fn_sig`, populate typed aliases, `canonicalize_named` sets `id` without rewriting `name`
- [ ] Verify locally: `cargo test --features wasm` and `aver compile examples/core/order_total.av --target wasm-gc`
- [ ] Confirm follow-up scope (constructors → `CtorId`, type-name maps → typed keys) belongs to a separate PR

## Deferred (not blocking phase B)

- Constructors still live in `extra_sigs` keyed by canonical string. They have `CtorEntry`s in `SymbolTable` already but aren't keyed by `CtorId` in the checker yet — migrating that is its own pass.
- `record_field_types`, `type_variants`, `value_members`, `opaque_types` stay string-keyed using canonical-form keys with on-the-fly resolution through `canonical_type_name`. Migrating each to `TypeId`-keyed is mechanical but heavy and out of phase B's stated scope.
- `pipeline::run` still rebuilds the symbol table after typecheck for `result.symbol_table`. Either remove the duplicate build or have typecheck stash its table into `result` — small follow-up.

### DIFF
--- a/aver-lsp/src/hover.rs
+++ b/aver-lsp/src/hover.rs
@@ -349,7 +349,7 @@ fn is_memo_safe_type(ty: &Type, safe_named: &std::collections::HashSet<String>) 
         Type::Tuple(items) => items.iter().all(|item| is_memo_safe_type(item, safe_named)),
         Type::List(_) | Type::Vector(_) | Type::Map(_, _) | Type::Fn(_, _, _) => false,
         Type::Result(_, _) | Type::Option(_) => false,
-        Type::Named(name) => safe_named.contains(name),
+        Type::Named { name, .. } => safe_named.contains(name),
         // Unresolved generics + recovery types — never memoise; treat
         // as opaque conservatively.
         Type::Var(_) | Type::Invalid => false,
--- a/examples/services/redis.av
+++ b/examples/services/redis.av
@@ -3,7 +3,7 @@ module Redis
         "Minimal Redis client over raw TCP using the RESP protocol."
         "Demonstrates persistent Tcp connections: connect once, issue multiple"
         "commands on the same socket, close explicitly."
-    exposes [connect, disconnect, ping, set, get, del]
+    exposes [connect, disconnect, ping, set, get, del, RedisConfig]
     depends []
     effects [Console, Tcp]
 
--- a/src/ast/types.rs
+++ b/src/ast/types.rs
@@ -133,17 +133,17 @@ impl Type {
                 // distinct `TypeId`, must stay incompatible).
                 (Some(a), Some(b)) => a == b,
                 // Exactly one side carries a `TypeId`: the typechecker
-                // already classified that side, the other side is a
-                // raw stamp (parser output / builtin / in-flight
-                // recovery). Phase B post-review tightens this branch:
-                // the pre-phase-B suffix fallback would have happily
-                // matched `Named { Some(A_Shape), "A.Shape" }` against
-                // `Named { None, "Shape" }`, re-introducing the
-                // cross-module collapse the typed identity is meant to
-                // prevent. Require strict-name equality so a Some-id
-                // side never matches a string that could canonicalise
-                // to a different ID.
-                (Some(_), None) | (None, Some(_)) => name_a == name_b,
+                // already classified that side; the other side
+                // attempted resolution and came up empty. Always
+                // reject — silent name fallback here was the round-6
+                // entry-fallback bug, where a dep module's
+                // unresolved bare `Shape` silently bound to the
+                // entry module's own `Shape`. Builtins like
+                // `HttpResponse` never set `id` on either side, so
+                // they exercise the `(None, None)` branch below;
+                // genuine cross-module typed/raw mixes hit this
+                // branch and must fail.
+                (Some(_), None) | (None, Some(_)) => false,
                 // Both sides unresolved (raw stamps, builtin records,
                 // tests). Keep the historical suffix relation as a
                 // best-effort match — there's no typed identity to
--- a/src/ast/types.rs
+++ b/src/ast/types.rs
@@ -132,12 +132,23 @@ impl Type {
                 // names happen to coincide (cross-module `Shape` —
                 // distinct `TypeId`, must stay incompatible).
                 (Some(a), Some(b)) => a == b,
-                // At least one side is unresolved (builtin record, in-
-                // flight stamp, parser output not yet through the
-                // checker's resolve pass). Fall back to the source-name
-                // suffix relation that has been the compatibility
-                // contract since before opaque IDs existed.
-                _ => {
+                // Exactly one side carries a `TypeId`: the typechecker
+                // already classified that side, the other side is a
+                // raw stamp (parser output / builtin / in-flight
+                // recovery). Phase B post-review tightens this branch:
+                // the pre-phase-B suffix fallback would have happily
+                // matched `Named { Some(A_Shape), "A.Shape" }` against
+                // `Named { None, "Shape" }`, re-introducing the
+                // cross-module collapse the typed identity is meant to
+                // prevent. Require strict-name equality so a Some-id
+                // side never matches a string that could canonicalise
+                // to a different ID.
+                (Some(_), None) | (None, Some(_)) => name_a == name_b,
+                // Both sides unresolved (raw stamps, builtin records,
+                // tests). Keep the historical suffix relation as a
+                // best-effort match — there's no typed identity to
+                // disagree with on either side.
+                (None, None) => {
                     name_a == name_b
                         || name_a.ends_with(&format!(".{}", name_b))
                         || name_b.ends_with(&format!(".{}", name_a))
--- a/src/ast/types.rs
+++ b/src/ast/types.rs
@@ -1,9 +1,11 @@
-/// Aver static type representation.
-///
-/// Lives under `crate::ast` so that `Spanned<T>` can carry an optional
-/// `Type` annotation without forming a cycle through `crate::types`
-/// (which depends on `crate::ast`). The original `crate::types::Type`
-/// is re-exported from here for backward compatibility.
+//! Aver static type representation.
+//!
+//! Lives under `crate::ast` so that `Spanned<T>` can carry an optional
+//! `Type` annotation without forming a cycle through `crate::types`
+//! (which depends on `crate::ast`). The original `crate::types::Type`
+//! is re-exported from here for backward compatibility.
+
+use crate::ir::TypeId;
 
 #[derive(Debug, Clone, PartialEq)]
 pub enum Type {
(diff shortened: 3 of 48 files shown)


jasisz:
All four review points addressed in c3de33cc:

1. **`finalize_check_result` last-write-wins** — the `HashMap<FnId, _>` → `HashMap<String, _>` export now detects bare-name conflicts (in deterministic `FnId` order) and skips the bare alias when two distinct FnIds share a bare name. Canonical `"Module.foo"` is always present.
2. **`bare_fn_aliases` / `bare_type_aliases` last-wins** — introduced `Resolution<_>` with `Single` / `Ambiguous`; `resolve_fn_id` / `resolve_type_id` return `None` on `Ambiguous`, forcing the user to qualify.
3. **Raw `Type::compatible` callsites** — `infer/expr.rs:990` and `infer/list_calls.rs:88` routed through `self.compatible`. Plus the raw matcher tightened: when exactly one side carries `id: Some`, strict-name equality only (no suffix fallback).
4. **Record stamps without TypeId** — `infer_record_create_expr` / `infer_record_update_expr` now return `self.canonicalize_named(Type::named(...))` instead of `Type::named(...)`.

Two regression tests added:
- `cross_module_same_bare_fn_name_drops_bare_alias_in_exported_map` — verifies the bare `doit` slot is absent from the exported map when two dep modules expose it.
- `cross_module_same_bare_type_name_bare_reference_is_ambiguous` — typed-identity rejection at the A.Shape vs B.Shape call site.

**Verified**: 1610/1610 tests pass, clippy `-D warnings` clean, fmt clean, `order_total.av --target wasm-gc` md5 byte-identical.

jasisz:
Round 2 review fixed in 515db796:

1. **(high) Ambiguous bare `Shape` typechecking** — the matcher's mixed `(Some, None)` branch was falling through to source-name equality, and an ambiguous bare alias produced `id: None` on the param side without a way for the matcher to know it was *deliberately* suppressed. Added `TypeChecker::type_name_is_ambiguous` and split the mixed-id case: ambiguous → reject, builtin/external → strict name equality (no suffix fallback). Verified with the corrected regression test below.

2. **(medium) Test name didn't match what it tested** — `cross_module_same_bare_type_name_bare_reference_is_ambiguous` now actually has `fn takesShape(s: Shape) -> Float` with `A.Shape.Circle(...)` and `B.Shape.Triangle(...)` call sites. Both reject.

3. **(medium) Entry-local fn over-suppressed** — `finalize_check_result` was treating any entry+dep bare collision the same as dep+dep. Now entry-scope owners always win the bare alias slot; the suppression only fires on dep-dep ambiguity when no entry owner exists. New regression test `entry_local_fn_shadows_dep_module_bare_alias` locks this in (Main.doit shadows Helper.doit; bare `doit` resolves to entry's `Int` return type, not Helper's `String`).

**Verified**: 1611/1611 tests pass, clippy `-D warnings` clean, fmt clean, `order_total.av --target wasm-gc` md5 byte-identical.

jasisz:
Round 3 fixes in 0b16cdee:

1. **Test strengthened** — `cross_module_same_bare_type_name_bare_reference_is_ambiguous` now asserts *exactly two* argument-mismatch diagnostics (filtered by `Argument 1 of 'takesShape'`) plus an explicit ambiguity diagnostic. An asymmetric regression (one side rejected, the other silently accepted) would fail the new count assertion.

2. **Explicit ambiguity diagnostic** — added `TypeChecker::ambiguous_type_candidates` + `report_ambiguous_named` and wired them into `build_signatures` (params + return types) and `register_type_def_sigs` (sum variants + record fields). Users now see:
   ```
   Function 'takesShape', parameter 's':
       Ambiguous type name 'Shape'; use `A.Shape` or `B.Shape` to disambiguate
   ```
   ahead of the downstream mismatch errors instead of just `expected Shape, got Shape`.

3. **CI red — `aver-lsp` leftover** — `aver-lsp/src/hover.rs:352` had a leftover `Type::Named(name)` tuple pattern. Fixed. Verified locally with the workspace clippy invocation CI runs: `cargo clippy --workspace --all-targets --exclude aver-lang -- -D warnings`.

**Verified**: 1611/1611 tests, workspace clippy clean, fmt clean, `order_total.av --target wasm-gc` md5 byte-identical.

jasisz:
Both nits closed in c2fa8e6d:

1. **Private candidates leaking into the diagnostic** — `Resolution::Ambiguous` now carries `Vec<TypeId>` (the actual visibility-exposed candidates populated via `merge_bare_type_alias`). `ambiguous_type_candidates` reads off that list — never scans the full `symbol_table.types` — so a dep module's `Shape` that isn't on its `exposes` list cannot show up in the user-facing "use `A.Shape` or `B.Shape`" suggestion. New regression test `ambiguity_diagnostic_omits_private_dep_candidates` locks this in with three dep modules (A, B exposing Shape; C with a private Shape).

2. **Binding annotations missing the diagnostic** — `x: Shape = ...` inside a flow body now goes through `canonicalize_named` + `report_ambiguous_named` in `flow.rs`. Also canonicalizes the bidirectional `parsed_ann` hint for consistency. New regression test `ambiguous_bare_name_in_binding_annotation_surfaces_explicit_diagnostic` covers it.

**Verified**: 1613/1613 tests, both clippy steps (workspace + aver-lang lib/bin) clean, fmt clean, `order_total.av --target wasm-gc` md5 byte-identical.

jasisz:
F1 closed in a0fc047c.

The core fix: the typechecker now has its own `visible_fn_ids` / `visible_type_ids` sets distinct from the symbol table itself. `SymbolTable` keeps its full dep-types-regardless-of-visibility shape (other consumers — proof IR, codegen — want that view), but `TypeChecker::resolve_*_id` filters every symbol-table look-up through the visible sets.

Population path:
- `insert_fn_sig` and `register_type_def_sigs` mark the current module's own declarations visible.
- `integrate_registry` marks every registry entry visible (registry is already filtered by `SymbolRegistry::from_modules` against the `exposes` contract).

Diagnostic:
- New `type_name_is_private_import` / `fn_name_is_private_import` distinguish "doesn't exist anywhere" from "exists but not exposed".
- `report_ambiguous_named` was generalized into `collect_unresolved_into` and now emits both ambiguity and private-import diagnostics from the same signature/binding hook.
- User sees: `Function 'takesPrivate', parameter 's': Type 'C.Shape' is not exposed by module 'C' — add 'Shape' to its `exposes` list to import it`.

Implementation note: `insert_fn_sig` had a chicken-and-egg with the visibility filter (it called `resolve_fn_id` on names that weren't yet visible). Solved with new `fn_id_for_canonical` / `type_id_for_canonical` helpers that skip the filter at the specific call sites that bootstrap visibility. Production look-ups still go through `resolve_*_id`.

New regression test: `qualified_private_dep_type_does_not_resolve` exactly mirrors your repro.

**Verified**: 1614/1614 tests, both clippy steps clean, fmt clean, `order_total.av --target wasm-gc` md5 byte-identical.

jasisz:
Both new repros fixed in 0fc5f405. CI on the previous commit went green too (all four checks passed) — the bug you caught was deeper than CI's gates.

The fix: new `canonicalize_named_in_module(ty, owner_module)` resolves bare names against the OWNER's symbol-table scope (`TypeKey::in_module(owner_module, name)`), bypassing the importer's `current_module_prefix` / `bare_type_aliases` and the visibility filter (owners always see their own types). `integrate_registry` uses it for every Function param/return, Constructor field, and RecordField. The resulting `TypeId` becomes the soundness anchor — even when the type itself isn't exposed, the matcher rejects cross-module collisions because distinct `TypeId`s never compare equal.

Two regression tests mirror the repros:
- `dep_module_exported_sig_resolves_in_owner_scope` — `B.make() -> B.Shape` returning through `takeB(s: B.Shape)` now typechecks cleanly when bare `Shape` is ambiguous (A.Shape vs B.Shape).
- `private_exported_sig_types_do_not_collapse_across_modules` — `D.consume(C.make())` is now rejected when C.Shape and D.Shape are both private but distinct.

1616/1616 tests, both clippy steps clean, fmt clean, `order_total.av --target wasm-gc` md5 byte-identical.

jasisz:
Round 6 fixes pushed in cd44e628. Took the architectural step the peer recommended (per-owner resolver context), not just patching the symptom.

**The three coordinated fixes:**

1. **Per-owner resolver context** — added `module_depends: HashMap<String, Vec<String>>` populated from each loaded module's declaration. `canonicalize_named_in_module` now resolves through three cases in the OWNER's context: qualified-with-visibility-check, bare-in-owner, bare-in-owner-depends. **No entry-scope fallback for dep modules** — an unresolved bare name in a dep stays unresolved instead of silently binding to the entry's own type.

2. **Sub-checker honors owner's depends** — `check_loaded_module_bodies` filters the sibling list passed to the sub-checker by the dep's own `depends [...]` declaration. The sub-checker for B no longer sees siblings B didn't import.

3. **Matcher strict on mixed-id** — the `(Some, None) | (None, Some)` branch now rejects outright (in both `Type::compatible` and the checker's matcher). Builtins exercise `(None, None)`; user-source typed/raw mixes are a resolution gap and must surface. The matcher also no longer auto-resolves the unresolved side via `resolve_type_id` — that was silently re-resolving in the importer's context.

Plus a complement: `infer_type` runs `canonicalize_named` on every expression stamp so `Spanned::ty` always carries `id` when the current checker can resolve, closing the matcher-auto-resolve gap upstream.

**Surfaced a real bug in `examples/services/redis.av`** — it was constructing `RedisConfig` from outside the module without exposing it. Added `RedisConfig` to Redis's `exposes` list (the example now matches the architectural intent).

Two new regression tests directly mirror your repros:
- `dep_module_resolver_uses_owner_depends_not_importer_siblings`
- `dep_module_resolver_does_not_fall_back_to_entry_types`

**Not** the full ResolvedAST/HIR resolve pass — that's still Phase E. But the *resolver split* (SymbolTable says what exists; per-owner resolver says what's visible from where) is now applied to types end-to-end. The matcher relies on canonicalise sites stamping `id` correctly in the right owner context, instead of re-resolving in whatever scope it happens to run in.

**Verified**: 1618/1618 tests, both clippy steps clean, fmt clean, `order_total.av --target wasm-gc` md5 byte-identical.

jasisz:
Round 7 fixed in abc2bfce. The bug was one line of intent: `canonicalize_named` was overwriting an existing `Some(TypeId)` whenever the source name happened to resolve in the current context. The new `infer_type` re-stamping wrapper from round 6 then turned that into a silent typed-identity collapse on every re-canonicalised expression.

Repro exactly as you described: `C.make()` correctly returned `Some(C.Shape)`, then Main's `infer_type` ran canonicalise in Main's context, the bare `Shape` resolved to A.Shape, and the canonicaliser blindly replaced `Some(C.Shape)` with `Some(A.Shape)`. Matcher saw `Some(A.Shape) == Some(A.Shape)` and accepted.

Fix: `Some(_)` is sacred. Both canonicalisers (`canonicalize_named` and `canonicalize_named_in_module`) only fill in `id: None`. Once an upstream boundary has stamped a typed identity in the correct owner context, no later pass can override it.

Regression test: `canonicalize_named_does_not_overwrite_existing_typeid` mirrors your repro exactly (public A.Shape, private-Shape-but-public-make C, Main forwards `C.make()` into `A.consume`).

**Verified**: 1619/1619 tests, both clippy steps clean, fmt clean, `order_total.av --target wasm-gc` md5 byte-identical.

PR #444, opened 2026-06-07
proof: accumulator-roundtrip proves on Lean + Dafny; recognize count field by role

jasisz:
## What

Makes the **accumulator-threaded encoder/inverse roundtrip** law prove on **both** proof backends, and removes the recognizer's brittle field-name coupling.

`examples/data/rle.av` carries `verify encode law roundtrip: decode(encode(xs)) => xs`. Naive head/tail induction can't close it — `encode = encodeLoop(xs, freshAcc)` threads an accumulator, so after peeling the head the induction hypothesis is about a *different* accumulator than the recursive call produces. The fix is the helper stack a human would write: generalize the loop over **all** accumulators satisfying a synthesized `count >= 0` well-formedness guard (the guard came from a counterexample — `count = -1` falsifies the fold step), then derive the user law at the initial accumulator.

## Commits

1. **`aa51162a` — Lean synthesis** (core, authored by a separate autonomous agent; clippy/fmt + verification only here).
2. **`af67008e` — Dafny port + name-agnostic recognizer** (this session's work):
   - **Dafny port.** The Dafny backend now emits the same proven support stack (repeat-succ, decode-append, WF-guarded fold step, count-nonneg, the generalized invariant) plus a `StringJoinEmptyChars` prelude lemma in `common.dfy` (the chars/join inverse — the Dafny counterpart of Lean's `String.intercalate_empty_chars`). The list roundtrip **and** its string wrapper now verify under Z3; deep concrete samples that exceed Dafny's evaluation fuel are discharged from the proven universal lemma instead of re-derived by unfolding. rle Dafny budget **3 errors → 0 errors, 0 axioms**.
   - **Name-agnostic recognizer.** The count field is now identified by its structural role (the field the step tests `== 0` and finish re-emits) instead of a hard-coded `"count"` name. A field-renamed clone of rle.av proves at `0/0` where it previously fell to `assume {:axiom}` — locked in by a regression test.

## Verification

- rle proves on Lean (`0 sorry`) **and** Dafny (`0 errors, 0 axioms`).
- Full `proof_spec` **72/72**, no regressions (the shared `common.dfy` additions are self-contained proven lemmas).
- `clippy -D warnings` clean, `fmt` clean.

## Not in this PR (follow-ups)

- Port the relaxation breadth beyond the RLE-shaped family (general accumulator invariant synthesis stays research-level).
- A general "samples ride the proven universal lemma" path (here it's gated to the two roundtrip strategies).

## Provenance

The Lean core (commit 1) was authored by a separate autonomous coding agent from a hand-off spec. The Dafny port, recognizer relaxation, and tests (commit 2) are this session's work. Flagged here at the author's request — not claimed as hand-authored by the repo owner.

### DIFF
--- a/src/codegen/dafny/mod.rs
+++ b/src/codegen/dafny/mod.rs
@@ -643,6 +643,41 @@ function StringJoin(sep: string, parts: seq<string>): string
   else parts[0] + sep + StringJoin(sep, parts[1..])
 }
 
+// The tail of StringChars(s) is StringChars of the tail of s. A
+// support lemma for StringJoinEmptyChars below.
+lemma StringCharsTail(s: string)
+  requires |s| >= 1
+  ensures StringChars(s)[1..] == StringChars(s[1..])
+{
+  assert |StringChars(s)[1..]| == |StringChars(s[1..])|;
+  forall j | 0 <= j < |s| - 1
+    ensures StringChars(s)[1..][j] == StringChars(s[1..])[j]
+  {
+    assert s[1..][j] == s[j + 1];
+  }
+}
+
+// Joining the single-character pieces of a string with the empty
+// separator rebuilds the string: StringJoin("", StringChars(s)) == s.
+// The proof-export inverse of the chars/join pair; the Dafny
+// counterpart of the Lean prelude's String roundtrip theorem.
+lemma {:fuel StringJoin, 2} StringJoinEmptyChars(s: string)
+  ensures StringJoin("", StringChars(s)) == s
+  decreases |s|
+{
+  if |s| == 0 {
+    assert StringChars(s) == [];
+  } else if |s| == 1 {
+    assert StringChars(s) == [[s[0]]];
+    assert s == [s[0]];
+  } else {
+    StringCharsTail(s);
+    StringJoinEmptyChars(s[1..]);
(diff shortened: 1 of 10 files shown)


jasisz:
Closing per the 'one generalization for both rle AND json, or it goes' criterion.

Verdict: NOT one generalization. rle is a fold-accumulator roundtrip (flat list, single tail-recursive worker, forward accumulator, concat invariant, unconditional). json/grok are recursive-descent prefix-parser roundtrips (recursive tree, mutual recursion across parseValue/parseArray/parseObject/parseString, position threading + suffix generalization, position arithmetic, AND conditional — floats don't round-trip without canonicalization, hence the jsonRoundtripSafe guard). They share only the trivial 'do induction generalizing the threading parameter' slogan; every concrete part (invariant, helper stack, induction structure, even unconditional-vs-guarded truth) differs.

So the accumulator-roundtrip strategy proves exactly the RLE shape (+ field-renamed clones) and nothing else in the corpus — a ~22-gate recognizer in the compiler for one toy example. Not worth keeping. The real unproven surface (16 json + 2 grok roundtrips on assume {:axiom}) needs the verified-parsing machinery, a separate and much larger effort.

Branch retained locally; reopenable if that effort ever subsumes this.

PR #536, opened 2026-06-16
wasm-gc: skip the bignum prelude for provably-bounded Int programs (whole-program gate)

jasisz:
etap 2, slice 2 — the first wasm-gc SIZE recovery from the unboxing analysis. Builds on slice 1 (#534, the MIR `Box`/`Unbox` rewrite). **No semantics change, fail-closed.**

## What it does
The wasm-gc emitter already has a complete scalar-`i64` Int path (used when `registry.bignum == false`). This PR narrows the gate that sets `bignum`: a program is **fully bare** — and takes the scalar-`i64` path with NO `$AverInt` slots and NO bignum prelude — iff the analysis proves **no node can produce an out-of-i64 value** (interval-checked) AND it uses no arbitrary-precision Int builtin (`String.fromInt`, `Int.fromString`, `Int.div/mod`, `Int.fromFloat`). Otherwise the whole program stays boxed (`Int = ℤ`, the prelude emitted). `rewrite_for_wasm_gc` shares the analysis + node-insertion core with `rewrite_for_rust`; the gate is `program_is_fully_bare`.

## Result
- `a*a` at i64::MAX (overflows → NOT bare) stays BOXED and exact: `85070591730234615847396907784232501249`. The C0 law does not regress.
- `countdown` (bounded loop) → bare i64, prelude DCE'd: **`--optimize size` 2778 B → 415 B**. `factorial` stays boxed (unbounded `acc`); `hello-worker` 577 B unchanged.
- The bare path equals the VM on bounded inputs (scalar i64 == ℤ, ring homomorphism).

## Honest scope limit (deferred to slice 3, fail-closed)
This is a WHOLE-PROGRAM gate, not per-slot. The win lands on pure-compute-returning-Int programs; any program that **stringifies an Int** (`String.fromInt` — the common "compute then print" shape) stays fully boxed, since that builtin can render a Big. The per-slot mixed case (a bounded value going bare while an unbounded one stays boxed, e.g. `factorial`'s counter; `String.fromInt(bounded)` using the i64 formatter) — which actually consumes the slice-1 `Box`/`Unbox` nodes in the wasm-gc body emitter — is the slice-3 follow-up. The `Box`/`Unbox` arms in the wasm-gc emitter remain `Ok(None)` placeholders, unreached (fully-bare → no boundary nodes; mixed → boxed).

## Verification
- `a*a` exact + boxed; `countdown` bare + correct + 415 B; bare path == VM on bounded inputs.
- `cross_backend_proptest` 16/16; `wasm_gc_spec` 37/37; effect-arg-overflow 4/4; full `cargo test --features wasm,wasip2` (38 binaries) green; lib 929 incl. 5 new gate tests; Rust slice-1 path unchanged (shared rewrite core). `cargo fmt --check` + `cargo clippy --features wasm,wasip2 -- -D warnings` clean.

### DIFF
--- a/src/codegen/wasm_gc/builtins/bignum.rs
+++ b/src/codegen/wasm_gc/builtins/bignum.rs
@@ -839,7 +839,7 @@ mod tests {
     /// `validation_guard` sibling module so its full-validate test renders
     /// helpers against an identical type section.
     pub(super) fn bignum_registry_for_validation() -> TypeRegistry {
-        let mut registry = TypeRegistry::build_with_handler(&[], &[], false);
+        let mut registry = TypeRegistry::build_with_handler(&[], &[], false, false);
         registry.bignum = true;
         registry.string_array_type_idx = Some(0);
         registry.aint_mag_array_idx = Some(1);
--- a/src/codegen/wasm_gc/module.rs
+++ b/src/codegen/wasm_gc/module.rs
@@ -139,9 +139,6 @@ pub(super) fn emit_module_with(
     let symbol_table = view.symbol_table;
     let resolved_fn_defs = view.resolved_fn_defs;
 
-    let mut registry =
-        TypeRegistry::build_with_handler(items, &resolved_fn_defs, handler_name.is_some());
-
     // Lower the post-link resolved fns to MIR and run the shared
     // `optimize` pipeline — the SAME six passes the VM consumes
     // (`ir::mir::optimize`). MIR is the only codegen path. Hoisted to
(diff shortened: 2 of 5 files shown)


jasisz:
Abandoning the whole-program gate: per-slot (consuming the slice-1 Box/Unbox nodes) subsumes it, and an adversarial panel found two C0 holes here — a gate-local one (Int.fromString slips the AP-builtin check) and a DEEPER one in the shared bare_i64.rs analysis (multi-tail-call counter under-approximation, which also affects the SHIPPED Rust backend). Fixing the shared analysis first, then doing per-slot.

PR #572, opened 2026-06-18
the-method: capability-fence the conjecturer (split proposer/runner)

jasisz:
## Problem

The Method's proposer was a single `general-purpose` agent with full tools — it could read the generated `.lean`, run `lake`, and search the web for tactics. "You are a conjecturer, not a prover" was prompt-intention with **nothing enforcing it**. The measured failure was exactly that: on a prover-blocked goal the model spiralled into debugging Lean tactics and the proof residual (the dominant token sink), because nothing physically stopped it.

## Fix — make it structural, not asked

Split the loop into capability-fenced roles (new custom agent types under `.claude/agents/`):

- **`the-method-proposer`** (`Read, Glob` only): the conjecturer. No Bash, no toolchain, cannot open any generated `.lean`/`.dfy`. It proposes Aver helper laws and nothing else — it *physically cannot* slide into tactic/residual debugging.
- **`the-method-runner`** (`Read, Write, Edit, Bash`): mechanically splices the proposed laws into a `/tmp` copy, runs `aver proof --discover` + `--check --check-json`, and returns **only the Aver-level verdict** (`universal`/`sorries`, which helper failed its bounded check). The Lean residual never crosses back to the conjecturer.
- **`the-method-verifier`** (`Read, Write, Edit, Bash`): the existing independent re-verify + persist-to-`decomposed/` gate, now its own restricted type.

One independent chain per task runs in parallel; within a chain the propose→run rounds refine against the Aver-level verdict only, up to the attempt cap. The optional `model` override now applies to the conjecturer alone; runner and verify gate stay on the default model.

## Scope / safety

- No Rust/codegen change — the `aver` toolchain and the proof corpus are untouched.
- `the-method-loop.js` rewritten to the split; `SKILL.md` updated to describe the fence.
- Statically validated (workflow script parses; agent files well-formed). Not yet live-run (invoking the workflow is an explicit, billed opt-in).

### DIFF
--- a/.claude/agents/the-method-proposer.md
+++ b/.claude/agents/the-method-proposer.md
@@ -1,0 +1,19 @@
+---
+name: the-method-proposer
+description: The Method conjecturer — proposes auxiliary Aver helper laws to unblock an OPEN Aver proof obligation. Read-only by design (no Bash, no toolchain, no access to any generated proof), so it physically cannot drift from conjecturing into tactic/prover debugging. Invoked by the the-method workflow.
+tools: Read, Glob
+---
+
+You are THE CONJECTURER in "The Method". Your one job: given an OPEN Aver `verify … law`, output the auxiliary Aver helper law(s) that would unblock it. You PROPOSE; a separate runner tests and the Lean kernel / Z3 judges. You never see the proof and never run anything — that is deliberate.
+
+Hard boundaries (these are enforced by your tool set, not just asked):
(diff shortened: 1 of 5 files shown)


jasisz:
Superseded by the full version (fence + drop --discover + aver-check pre-flight + bridge prompt + model tiering + 2 verified closures). See the new PR.

PR #700, opened 2026-07-11
Decline non-list parametrized constructors fail-closed

jasisz:
Follow-up hardening to the list-cons leg (#697). The element-type binding that pins a list constructor's element type to the byte-derived type section was gated on the literal type name `List`; a future parametrized non-list constructor (any `.app1`/`.app2` result that is not `List<_>`) would skip that binding and certify via the struct index alone, reopening the round-1 relabel hole for that ADT if its field were erased.

## Change

- New guard `parametrizedConstructIsList`: a parametrized constructor result may only be certified when it is the recognized list cons shape (whose element type is byte-bound). Any other parametrized result is declined fail-closed until a per-shape binding exists.
- Monomorphic constructors are unaffected — they stay bound by the struct index carried in the code entry.
- Witness generators (both emitters) and the guard-iso test witnesses emit the matching `by decide` proof of the new conjunct.
- Certificate schema version 41 -> 42.

## Not exploitable today; this is fail-closed hardening

Verified: the only parametrized-result construct claims in the entire corpus are `.app1 "List"` (singletonJsonEntries, prependJsonEntry), which satisfy the new guard. No non-list parametrized constructor is certified anywhere, so nothing is dropped. Found by self-review during #697 and banked as probe-artifacts/construct-narrowing-nit; this closes it.

## Tests

fmt, debug build, scoped clippy -D warnings, lib 1019, cert_certify_spec 11, cert_decode_spec 2, cert_verify_spec 29; the honest json certificate still certifies 12 exports and cert_goals stays 23/0.

A guard-iso adversary for a non-list parametrized result was skipped for lack of a fixture supplying the byte witnesses (no such export exists); the fail-closed behavior follows by construction from the conjunct.

### DIFF
--- a/src/codegen/cert/AcceptedArtifactCore.lean
+++ b/src/codegen/cert/AcceptedArtifactCore.lean
@@ -366,6 +366,18 @@ def isListConstructSymPlan (symPlan : SymRawPlan) : Bool :=
   | .app1 "List" _ => true
   | _ => false
 
+/-- A parametrized constructor result may only be certified when it is the
+    recognized `List` cons shape (whose element type is byte-bound below).
+    Any other parametrized result (`.app1 _`/`.app2 _`) is declined
+    fail-closed until a per-shape type binding exists; monomorphic results
+    stay bound by the struct index in the code entry. -/
+def parametrizedConstructIsList (symPlan : SymRawPlan) : Bool :=
+  match symPlan.result with
+  | .app1 "List" _ => true
+  | .app1 _ _ => false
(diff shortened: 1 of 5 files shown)


jasisz:
Superseded by #701: nominal sum-type roots make the sum carrier a concrete reference to the sum's root struct, which removes the representation gap this guard tried to patch. The guard added here was also bypassable through named constructor forms, so it never reached a shippable shape. Closing without merge; the branch stays parked for reference.

PR #736, opened 2026-07-16
Report modules without the Int runtime helper as uncertified instead of failing --certify

jasisz:
## Problem

`aver compile --certify` exited with code 1 and an internal error (`module has no __rt_aint_from_i64 box helper`) on any module that never boxes big integers, such as `examples/core/hello.av`. The CHANGELOG promises that unsupported exports remain explicitly uncertified with a reason, so a module without integer arithmetic must still produce its certificate package instead of aborting the compilation.

## Root cause

The certificate disassembler treated the `__rt_aint_from_i64` export as mandatory and returned a whole-module error when it was missing. The fix makes the helper's index `Option<u32>` throughout the engine: recognizers for integer-shaped functions decline to certify when the helper is absent (strictly fewer bodies match, so nothing new can certify), and each affected export is declared uncertified with a readable reason naming the missing Int carrier support.

## Testing

- `cargo build` and `cargo build --bin aver --features wasm`: clean.
- `cargo test -p aver-cert`: 19 passed, including a new engine unit test covering analysis of a module without the helper export.
- `cargo test --test cert_certify_spec --features wasm certify_emits_declared_uncertified_package_for_module_without_int_helper`: the new integration test passes; it compiles `examples/core/hello.av` with `--certify` and asserts exit 0, the certificate package on disk, every export certified or declared uncertified with a non-empty reason, and no whole-module abort message.
- `./target/debug/aver compile --target wasm-gc --certify examples/core/hello.av -o <tmp>`: exit 0, prints the wasm artifact plus `certificate → cert/ (0 certified, 3 source-level-only)`.
- `./target/debug/aver compile --target wasm-gc --certify examples/certification/add_one.av -o <tmp>`: exit 0, `1 certified: addOne` — happy path unchanged.
- `cargo clippy -p aver-cert` and `cargo clippy --features wasm,wasip2 --tests`: no new warnings; `cargo fmt`: no drift.


jasisz:
Round 2 addresses the review finding that the certificate package emitted for a module without the Int runtime failed its own verification: the byte decoder resolves the module-wide host-role table to `none` for a module with no Int carrier type, while the producer declared a concrete all-null table, so the generated `rfl` pin could never close.

What changed:

- The manifest now represents the host-role table like it already represents the optional start function: a module without the Int carrier declares no table at all (`hostRoleTable := none` in `Manifest.lean`, `"hostRoleTable": null` in `cert-manifest.json`). A module with the carrier keeps the exact concrete table as before, wrapped in `some`.
- The acceptance pin now equates the decoder's full optional result with the declared manifest value. The equality is unconditional and identical in strength for both module classes: a carriered module must declare exactly the decoded table, and a carrierless module must declare the absence. Tampering either way is rejected — I confirmed empirically that editing a carrierless package to claim the old all-null table makes `aver cert verify` fail at exactly the host-role pin, and the whole-module guard isolation test now also proves a carriered artifact cannot claim an absent table.
- Claim matching treats an absent table as binding no host roles, so any claim citing a box/add/mul/sub role stays fail-closed.
- This is a certificate format change, so the manifest schema version is now 2 and the embedded wall identity was updated per convention.

End-to-end results on this machine:

- `examples/core/hello.av` (no Int runtime): the package now passes the complete Lean acceptance, including the fresh replay. The CLI still exits nonzero with `NO CERTIFIED EXPORTS (admission only, no behavioral claims)` because hello has zero certified exports and the deliberate policy (pinned by `empty_cert_is_admission_only_and_exits_nonzero`) is that an empty certificate never reads as a green verdict. The previous `DECLINED ... rfl failed` is gone; I did not weaken that exit-code policy.
- `examples/certification/add_one.av`: certify reports 1 certified, `aver cert verify` prints CERTIFIED and exits 0. A new regression test pins the add_one certification output (certified entry, contracts, and the concrete host-role table) to its pre-change shape.

New tests: the hello-style test now asserts the exact human-readable decline reasons (so a future false-positive classification fails loudly), and an adversarial engine fixture exercises the carrier-present/helper-absent corner across the recursion, dispatch, and structural classify paths.

PR #758, opened 2026-07-29
Give the shared arithmetic helpers low function indices

jasisz:
The notepad example now reaches CERTIFIED with twelve exports — the same twelve its Data.Json dependency certifies when compiled on its own.

The compiler assigned every builtin a function index after every user function, so in a program with 155 user functions the four shared bignum sub-routines landed at 166-169. The trusted wall compares the emitted helper bodies against byte templates that splice a call target as a single raw byte, which is only valid below 128 — above it wasm encodes the index as multi-byte LEB128, the synthesized body is a byte short, and the comparison correctly fails. The four sub-routines now get a low block immediately after the entry function, so they sit at 1-4 regardless of program size. Nothing forced the previous order: wasm function order is free, the verifier derives helper roles from the bytes and the export section, and certification conflicts with the optimizer so the emitted layout is what gets certified.

Measured both ways on a synthetic 151-function module: with the old layout the artifact proof reports that the arithmetic table check is false; with the new one it certifies. A new producer test pins the four indices below 128 and consecutive from the first user slot, identifying the block by its call-signature multiset rather than by position. A new isolation test in the guard suite repoints one call target inside the add helper and shows the table check flips to false while the honest control stays true.

Also included: the artifact proof header now sets an elaboration budget matching the one the certificate proof already had. This is precautionary, not a measured fix — no artifact in the corpus was observed to exhaust the stock allowance, and nothing in this branch stopped failing because of it; the comment and commit message say so.

The mutual code-table isolation probe previously hard-coded the two member indices; it now reads them from the export section of the module it just compiled. The four assertions, their projections and their kernel discharges are unchanged. That test failed loudly on stale literals rather than passing weakly, which is evidence its content survives the change; the numbering itself is pinned explicitly by the new producer test and the package snapshot.

Snapshot delta (five lines) is entirely implied by the reorder: the single user function moves from index 1 to 5 and the arithmetic parameters shift from 2-5 to 1-4. Every code entry byte, including both call immediates, and the whole host role table are unchanged. Wall assets untouched; no wall identity bump.

### DIFF
--- a/aver-cert/src/engine/render_project.rs
+++ b/aver-cert/src/engine/render_project.rs
@@ -1897,6 +1897,14 @@ fn render_artifact(
          -- The whole-module big-Nat closure fold is kernel reduction. This\n\
          -- explicit depth budget affects kernel reduction limits only, not soundness or axioms.\n\
          set_option maxRecDepth 200000\n\
+         -- Companion budget for the elaborator, matching the one Certificate.lean\n\
+         -- already carries. Precautionary rather than a fix for an observed\n\
+         -- failure: this proof is one fixed shape repeated per claim, so its\n\
+         -- elaboration cost grows with the size of the artifact, and the stock\n\
+         -- allowance is the only thing here that was never stated explicitly.\n\
+         -- Like the depth budget above it moves a resource limit only: no axiom,\n\
+         -- no trusted hypothesis, no change to what the kernel must accept.\n\
+         set_option maxHeartbeats 1600000\n\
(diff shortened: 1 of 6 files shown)


jasisz:
Superseded: the arithmetic template indices are being encoded properly in the trusted wall instead. Reordering the emitter's function indices only avoids the bound for function indices, leaves the carrier and limb type-index bounds biting from the other side, and would make the wall's multi-byte path inert — no real program would exercise it. The elaboration-budget commit carries over to that branch.

PR #762, opened 2026-07-29
Certify string concatenation in modules without an integer carrier

jasisz:
The hello-world example now carries its own proof: greet and shout certify and the kernel replays them.

They always matched the certified concatenation template byte for byte. What stopped them was unrelated to meaning: a module with no integer arithmetic anywhere never gets a carrier struct, so its function bodies declare no locals, while the wall's byte template hard-coded a locals prelude holding one carrier-typed local. Appending a trivial integer function to a copy of the file made both certify with no wall change at all, which is what identified this as an environmental dependency rather than a semantic one.

The template variant is chosen by a value the wall recomputes from the bytes, never by the package. A new strict three-state decoder reads the type section: it does not decode, it decodes with no carrier struct, or it decodes with the carrier at a given index. The claim declares that state and acceptance conjoins the recomputation, the locals prelude is a function of the pinned state alone, and the decoded locals vector of the real code entry must match the count that state implies.

One change reaches wider than the concatenation column and deserves the closest reading: obligations previously pinned their carrier against a decoder that returns nothing for a carrier-less module, so no such declaration was satisfiable. They now pin against the three-state reading with the carrier-less state mapped to a reserved index. On any module that has a carrier struct this is the identical statement, and a type section that does not decode still admits nothing.

The isolation test builds a two-by-two of one module — honest carrier struct versus a decoy differing in one storage byte, crossed with the one-local and zero-local preludes. Both honest states are accepted, both hostile states are rejected at the carrier-state conjunct, and a literal copy of the check with that one conjunct removed accepts both hostile artifacts. The attribution is an acceptance, not an inequality, so it cannot go vacuous under drift. Two further examples show each hostile module genuinely carries the code entry its claim synthesizes, so byte equality alone is blind to the attack.

Corpus sweep before and after changes exactly one line: the hello example goes from zero to two. Nothing else gained, nothing lost; the package snapshot drifts by the wall identity line only. Wall identity bumped; format and schema versions unchanged.


jasisz:
Updated after a three-lens panel refuted the inertness argument for the reserved carrier index.

The panel showed the arithmetic table check never reads the type section, so a module whose type section holds no wall-visible carrier struct could still present a role table carrying only the box helper, pass the role bound, and — with the reserved index supplying the last missing value — take claims from every family that needs a host role. The construction is realizable rather than hypothetical: the carrier decoder requires the third field to be an i32, but a packed i8 field decodes to a different tag, so a fully working carrier can be invisible to it.

The fix conjoins the byte-derived carrier state with the index the helper bodies splice, so an admitted arithmetic table now implies a carrier struct at exactly that index. Families that cite no arithmetic role are untouched, which is what keeps string concatenation certifying without a carrier. The reserved-index helper that collapsed the three-state decoder back to two values is deleted in favour of an explicit three-arm reading, so the carrier-less state and index zero stay distinguishable.

The new isolation test is built on the exploit itself: two assemblies of one real integer runtime differing in exactly one byte, with byte-identical box helpers, pinning the whole chain as facts before showing the real check rejects the invisible-carrier module while a copy with that one conjunct removed accepts both.

The anti-false-positive tripwire that fired in CI was resolved deliberately: it asserted the hello example must not certify, which is precisely what this change reverses. It is renamed, keeps every carrier-less manifest assertion, keeps an explicit assertion that no integer-family class appears, and now records the certification as the expected fact.

jasisz:
Second round of panel findings closed.

A second gap of the same class was found one family over: the compensating relaxation of the carrier binding had been applied to all ten non-expression families, while only string concatenation needed it. Four families reach acceptance without citing any arithmetic role, and one of them — the named-ADT constructor — carries an integer-representation face, so it could have been stated over a representation index the module provably has no carrier struct at. The relaxation is now wired to the string-concatenation claim list alone; every other family keeps the strict binding it had before this branch.

Both prose claims that asserted this could not happen were wrong and are corrected. The real criterion is not that a family cites no arithmetic role — four of them do not — but that its face never reads the carrier specification. The wall docstring and the format specification now say that, name the constructor family as the counterexample, and carry a must-not for reimplementors.

The isolation test gained an arm built on the exploit: in the invisible-carrier module the strict binding rejects the obligation both at its real index and at the reserved one, while the carrier-free binding accepts the reserved index — same artifact, same obligation, the flip attributable to the scoping alone. A truncated-module witness now covers the undecodable-type-section arm, which previously had none. The one-byte claim is now backed by a length assertion rather than resting on an unstated premise, the stub obligation carries a literal code table pinned to the decoder on both assemblies so both attributions are of equal strength, and the test's lakefile roots were reduced from the whole wall to the closure it actually needs, taking it to seventeen seconds.

The second deliberate tripwire that CI caught — the one pinning this module to an admission-only verdict — is resolved the same way as the first: the obsolete half updated, the assertion that the package must not be declined kept and strengthened by a positive green verdict, and an added assertion that the module still has no box helper so the test cannot silently stop being about a carrier-less module. Admission-only coverage itself is unchanged and still exercised end to end by the empty-certificate test on a fixture that certifies nothing.

jasisz:
Final round: the prose this branch added overclaimed and is corrected.

Four passages asserted that every family except string concatenation keeps the strict carrier binding, and that a module without a decodable carrier struct can carry string-concatenation claims and nothing else. Both are false about the expression-fragment family, which is deliberately absent from that predicate entirely and therefore satisfies neither binding. A fifth instance of the same wording was found while fixing the four. All are now qualified to non-expression-fragment families, with the reachability spelled out: the arithmetic arm for a module without helpers needs only the helper-absent fact, the role bound is vacuous on an empty table, and generic and projection fragments cite no role, so a carrier-less module can carry expression-fragment claims too.

Added in their place is an explicit statement of what actually binds an expression-fragment claim's carrier: byte equality alone. The index is spliced into the synthesized locals prelude and every carrier-typed immediate, the binding requires the result to equal the real code entry, and the integer-family renders additionally pin the declared function type. It is never tied to the decoded carrier state or to the arithmetic parameters. The docstring ends with a standing instruction not to describe this file's bindings as covering every family.

No definition changed in this pass. The gap that motivated the correction — the expression-fragment family having no byte-derived carrier binding while its faces model the box role at the claim's own carrier — predates this branch and is recorded for separate work with its own review.

Issue #753, opened 2026-07-28
Binary TCP support and SHA256 support.


n1bor:
# Binary protocol support: 4 blockers preventing a Bitcoin P2P client in Aver

**Environment:** `aver 0.27.1`, repo at commit `f581bc27`, Linux (WSL2).

Happy to split this into 4 separate issues if you'd prefer — say the word and I'll
file them individually and link back here. Keeping them together for now because
they share one root cause and one motivating use case.

## What I'm trying to build

I was having a play with Aver. So thought I would try implementing something "real". Most of below is analysis by Claude but looks accurate to me!

A command-line program in Aver that:

1. Takes the IP address of a Bitcoin node as a parameter
2. Connects to that node over the Bitcoin P2P protocol (TCP, default port 8333)
3. Listens for transaction announcements
4. Decodes each transaction and prints its details to the screen

This is a good fit for Aver on paper — transaction decoding is pure, total,
recursive parsing over a well-specified binary format, which is exactly the kind
of code `verify` blocks and refinement types shine on. The blocker is entirely in
getting the bytes in and out, plus one missing primitive.

The Bitcoin P2P wire format is a length-prefixed binary framing:

```
magic (4 bytes) | command (12 bytes, ASCII null-padded) | length (4 bytes LE) |
checksum (4 bytes) | payload (`length` bytes, arbitrary binary)
```

Mainnet magic is `F9 BE B4 D9`. The checksum is the first 4 bytes of
double-SHA-256 over the payload.

## Issue 1 — `Tcp.send` silently corrupts binary responses

**File:** `aver-rt/src/tcp.rs:129`

```rust
Ok(String::from_utf8_lossy(&buf).into_owned())
```

Every byte sequence in the response that isn't valid UTF-8 is replaced with
U+FFFD REPLACEMENT CHARACTER. The replacement is lossy and irreversible — the
original bytes cannot be recovered from the returned `String`.

For Bitcoin specifically, the first four bytes of every single message
(`F9 BE B4 D9`) are not valid UTF-8, so corruption begins at offset 0 of the
first read.

The failure is silent. `Tcp.send` returns `Result.Ok` with mangled data rather
than surfacing an error, so a caller has no way to detect that it happened.

**Applies to any binary protocol**, not just Bitcoin — this equally breaks
Redis RESP with binary payloads, MQTT, Postgres wire protocol, DNS over TCP, etc.
## Issue 2 — `Tcp.readLine` cannot read binary at all

[...]

PR #764, opened 2026-07-29
Add Tcp.sendBytes — a byte-clean one-shot TCP call

### LINKED ISSUE #763: `Tcp.send` silently corrupts non-UTF-8 responses, making binary protocols unreachable
Split out of #753 (Issue 1 of the four described there) so that a fix can close
this one on its own. The other three blockers in #753 are unaffected and remain
open.

## What happens

`Tcp.send` returns `Result.Ok` containing corrupted data whenever the peer's
response is not valid UTF-8. The corruption is silent and irreversible.

`aver-rt/src/tcp.rs:130` ends the read path with:

```rust
Ok(String::from_utf8_lossy(&buf).into_owned())
```

`from_utf8_lossy` replaces every byte sequence that isn't valid UTF-8 with
U+FFFD REPLACEMENT CHARACTER. Because 128 of the 256 possible single-byte
values — the whole `0x80`–`0xFF` range — all collapse to that same character,
the mapping is many-to-one and no inverse exists. Given a `�`, the byte that
produced it is unrecoverable.

## What I expected

Either the bytes arrive intact, or the call reports that it could not represent
them. What actually happens is neither: the data is destroyed and the call
reports success.

Two properties make this hard to notice:

- The return shape is `Result.Ok`, so no error path is taken.
- `String.len` counts Unicode scalar values, so a corrupted 12-byte read reports
  length 12 — exactly what a clean read would report. A length check does not
  catch it.

## Steps to reproduce

1. Start a TCP server that responds with four non-UTF-8 bytes. `F9 BE B4 D9`
   is a good probe — the Bitcoin mainnet magic, and each byte is invalid UTF-8
   for a different reason (`F9` can never begin a sequence under RFC 3629, `BE`
   and `B4` are orphaned continuation bytes, `D9` is a 2-byte leader whose
   continuation never arrives):

   ```python
   import socket, threading
   srv.bind(("127.0.0.1", 18334)); srv.listen(5)
   def handle(c):
       while c.recv(4096): pass
       c.sendall(bytes([0xF9, 0xBE, 0xB4, 0xD9]))
       c.close()
   while True:
   ```

2. Read from it with `Tcp.send`:

   ```aver
   module Repro
       intent = "Tcp.send against a binary responder."
       exposes [main]
       effects [Tcp.send, Console.print]

   fn main() -> Unit
       ? "Print whatever the peer sent."
       ! [Tcp.send, Console.print]
       Console.print("{Tcp.send("127.0.0.1", 18334, "hi")}")
   ```

3. Observe:

   ```
   Result.Ok("����")
   ```

   The four bytes are gone. Each `�` occupies 3 bytes (`EF BF BD`) in the
   resulting string, so the 4-byte response becomes 12 bytes of tombstone.

## Why this matters

It closes off binary protocols entirely. Bitcoin P2P is the case I hit —
mainnet framing begins `F9 BE B4 D9`, so corruption starts at offset 0 of the
first read and there is no message a node can send that survives. The same
applies to Redis RESP with binary payloads, MQTT, the Postgres wire protocol,
DNS over TCP, and anything else that is length-prefixed rather than
line-delimited.

There is no workaround from Aver source: the decode happens in the runtime,
before any Aver code sees the data.

## Related, but separate

Two adjacent gaps in the same area, described as Issues 2 and 3 of #753. They
have different causes and different fixes, so they are not covered here:

- `Tcp.readLine` cannot read binary at all — it frames on `\n` (wrong for
  length-prefixed protocols) and `BufRead::read_line` hard-errors on non-UTF-8
  rather than corrupting. Same underlying assumption, opposite symptom.
- `Tcp.writeLine` cannot *send* arbitrary bytes — it appends `\r\n`
  unconditionally, and its `String` parameter is UTF-8, so the byte `0xF9`
  cannot be expressed at all (it would go out as `C3 B9`).

Fixing this issue alone makes binary *responses* readable on the one-shot path;
the session path needs the other two.

## Proposed fix

An additive byte-clean sibling rather than a change to `Tcp.send`:

```
Tcp.sendBytes : (String, Int, List<Int>) -> Result<List<Int>, String>
```

Reshaping `Tcp.send`'s return type would break every existing caller, and making
it error on invalid UTF-8 would regress callers who currently tolerate lossy
text. A sibling leaves existing code untouched and lets the docs steer new code
to the correct one.

`List<Int>` follows the existing `Byte` namespace convention of operating on
`Int` rather than introducing a byte primitive, though `Vector<Int>` would give
O(1) indexed access, which matters when parsing a length-prefixed header.

Happy to be told a different shape is preferred.

n1bor:
# Add `Tcp.sendBytes` — a byte-clean one-shot TCP call

Fixes #763. Broader context, including the other three binary-protocol
blockers this does *not* address: #753.

## The problem

`aver-rt/src/tcp.rs` ends `send` with:

```rust
Ok(String::from_utf8_lossy(&buf).into_owned())
```

Every non-UTF-8 sequence in the response becomes U+FFFD. The substitution is
many-to-one — 128 of the 256 possible single bytes collapse to the same
replacement character — so it cannot be undone. It is also silent: the call
returns `Result.Ok` with mangled data, and `String.len` reports the same length
an uncorrupted read would, so neither the return shape nor a length check
reveals it.

That makes every binary protocol unreachable from Aver. Demonstrated against a
server that always replies with the four bytes `F9 BE B4 D9` (the Bitcoin
mainnet magic — chosen because each byte is invalid UTF-8 for a different
reason):

```
Tcp.send      -> Result.Ok("����")
Tcp.sendBytes -> Result.Ok([249, 190, 180, 217])
```

## The change

Adds `Tcp.sendBytes : (String, Int, List<Int>) -> Result<List<Int>, String>`.

Identical socket behaviour to `Tcp.send` — open, write, `shutdown(Write)`, read
to EOF, same 10 MB cap, same timeouts, same port validation — but the payload
and the response stay bytes end to end, with no encoding or decoding in either
direction.

### Design decisions

**Additive, not a fix in place.** Changing `Tcp.send`'s return type would break
every existing caller, and making it error on invalid UTF-8 would regress
callers who currently tolerate lossy text. A sibling method leaves existing code
untouched and lets the docs steer new code to the right one.

**`List<Int>` as the byte carrier.** Follows the existing `Byte` namespace
convention, which already operates on `Int` rather than introducing a byte
primitive. Worth a maintainer opinion: `Vector<Int>` would give O(1) indexed
access, which matters more when parsing a length-prefixed header than when
sending one. I went with `List<Int>` for consistency with `Args.get` and
`Disk.listDir`, and because `Vector.fromList` makes the conversion cheap at the
one place it matters.

**Out-of-range bytes are catchable, not fatal.** A value outside `0..=255`
returns `Result.Err("Tcp.sendBytes: byte 256 at index 1 is out of range
(0–255)")` rather than trapping — the same treatment the port range gets after
Phase 4.7+ fix #13. The index is included so a long payload is debuggable. A
non-`List<Int>` argument is still a type error, caught statically.

## What's wired

- `aver-rt::tcp::send_bytes` — the runtime call
- VM path: service dispatch, `register`/`register_nv`, `DECLARED_EFFECTS`,
  `effects`, and the `VmBuiltin` enum
- Effect classification: `GenerativeOutput`, with two new `RuntimeType`
  variants (`ListInt`, `ResultListIntStr`). Oracle `verify <fn> trace` works
  with a `given tcp: Tcp.sendBytes = [stub]` binding — confirmed against
  `.result`, `.trace.length()` and `.trace.contains()`

## What's not wired — deliberately, and I'd like direction

**Codegen backends.** `Tcp.sendBytes` is VM-only. I stopped there rather than
guess at three backends in a first PR:

- **Rust codegen** currently reports `Compiled ... [Rust]`, then the generated
  project fails at `cargo build` with "MIR walker could not render the `main` fn
  body". Not a silent miscompile, but the error arrives one stage too late. The
  binding itself looks like a short arm next to the existing `Tcp.send` one; the
  `Vec<u8>` ↔ `AverList` conversion is the only real work.
- **wasm-gc** rejects it — but so does any program with a `List<Int>` literal,
  with or without this change (`List literal: cannot resolve list instantiation
  (got List<Int>)`). I verified that against the released 0.27.1 binary too, so
  it's pre-existing and orthogonal.
- **wasip2** already rejects all `Tcp.*` as host territory, so nothing to do.

Happy to extend to any of these if you'd point me at the shape you'd want.

**CHANGELOG.** Left alone on purpose — `AGENTS.md` is explicit that
`tools/release.py` owns version headers, and adding an entry means guessing the
next version number.

`cargo fmt --all -- --check` passes clean (rustfmt 1.9.0) with no diff.

## Tests

Two added to `tests/eval_spec.rs`, alongside the existing `tcp_tests`:

- `tcp_send_bytes_round_trips_non_utf8` — echo server, asserts
  `[249, 190, 180, 217]` survives intact. Marked `#[ignore]` like its
  `tcp_send_echo_server` neighbour since it binds a socket.
- `tcp_send_bytes_rejects_out_of_range_byte` — asserts the error names both the
  offending value and its index.

Manually verified: empty payload → `Result.Ok([])`; connection refused,
out-of-range port, and a `String` passed where `List<Int>` is expected all
produce the right diagnostics.

### DIFF
--- a/aver-rt/src/tcp.rs
+++ b/aver-rt/src/tcp.rs
@@ -130,6 +130,35 @@ pub fn send(host: &str, port: i64, message: &str) -> Result<String, String> {
     Ok(String::from_utf8_lossy(&buf).into_owned())
 }
 
+/// Byte-clean sibling of [`send`].
+///
+/// Identical socket behaviour — open, write, `shutdown(Write)`, read to EOF —
+/// but the payload and the response stay `Vec<u8>` end to end. `send` converts
+/// the response with `String::from_utf8_lossy`, which replaces every non-UTF-8
+/// sequence with U+FFFD and cannot be undone; that makes it unusable for binary
+/// protocols whose framing bytes are not valid UTF-8. This function performs no
+/// encoding or decoding, so the caller sees exactly what the peer sent.
+pub fn send_bytes(host: &str, port: i64, payload: &[u8]) -> Result<Vec<u8>, String> {
+    validate_port(port)?;
+    let socket_addr = resolve(&format!("{}:{}", host, port))?;
+    let mut stream =
+        TcpStream::connect_timeout(&socket_addr, CONNECT_TIMEOUT).map_err(|e| e.to_string())?;
+    stream.set_read_timeout(Some(IO_TIMEOUT)).ok();
+    stream.set_write_timeout(Some(IO_TIMEOUT)).ok();
+    stream.write_all(payload).map_err(|e| e.to_string())?;
+    stream.shutdown(std::net::Shutdown::Write).ok();
+
+    let mut buf = Vec::new();
+    Read::by_ref(&mut stream)
+        .take(BODY_LIMIT as u64 + 1)
+        .read_to_end(&mut buf)
+        .map_err(|e| e.to_string())?;
+    if buf.len() > BODY_LIMIT {
+        return Err("Tcp.sendBytes: response exceeds 10 MB limit".to_string());
+    }
+    Ok(buf)
+}
+
 pub fn ping(host: &str, port: i64) -> Result<(), String> {
     validate_port(port)?;
     let socket_addr = resolve(&format!("{}:{}", host, port))?;
--- a/docs/services.md
+++ b/docs/services.md
@@ -257,6 +257,7 @@ Source: `src/services/tcp.rs`
 | Function | Signature |
 |---|---|
 | `Tcp.send` | `(String, Int, String) -> Result<String, String>` |
+| `Tcp.sendBytes` | `(String, Int, List<Int>) -> Result<List<Int>, String>` |
 | `Tcp.ping` | `(String, Int) -> Result<Unit, String>` |
 
 **Persistent connections:**
@@ -271,6 +272,15 @@ Source: `src/services/tcp.rs`
 `Tcp.Connection` is **opaque** from the surface: construction is reserved to `Tcp.connect` and field reads / pattern matches are rejected by the type checker. The handle is purely an identity token — the caller has nothing to inspect inside it. The underlying socket lives in a thread-local `HashMap` (VM / self-host / wasm-gc-bridge, keyed by `AtomicU64` "tcp-N") or a 256-slot wasm-gc array (`--target wasip2`, slot allocated via first-free scan + monotonic counter generation). Either way, manually forging an id is impossible: the type checker rejects the constructor.
 
 `Tcp.send` is stateless and ephemeral — it opens a fresh socket, writes the request bytes raw (no `\r\n` append), `shutdown(Write)` to signal end-of-request, then reads the peer's response until EOF, capped at 10 MiB. It does **not** touch the persistent-connection pool, so a program holding 256 live `Tcp.connect` handles can still issue `Tcp.send` to another peer. Stream errors (`stream-error.last-operation-failed`) surface as `Result.Err("tcp: stream error")`; a clean half-close (`stream-error.closed`) returns whatever the peer flushed.
+
+`Tcp.sendBytes` is the byte-clean form of `Tcp.send`: same socket behaviour, but
+the payload and response stay `List<Int>` and no UTF-8 encoding or decoding
+happens in either direction. Prefer it for any binary protocol. `Tcp.send`
+decodes the response with `String::from_utf8_lossy`, which replaces every
+non-UTF-8 sequence with U+FFFD — silently, irreversibly, and starting at the
+first offending byte — so it is only safe for protocols whose responses are
+valid UTF-8 text. Payload values outside `0..=255` return
+`Result.Err` naming the offending value and its index.
 
 ### `Random` namespace — use granular effects (`! [Random.int]`, `! [Random.float]`)
 
--- a/src/services/tcp.rs
+++ b/src/services/tcp.rs
@@ -282,9 +282,18 @@ fn bytes_arg(val: &Value, method: &str) -> Result<Result<Vec<u8>, String>, Runti
     let mut out = Vec::with_capacity(items.len());
     for (idx, item) in items.iter().enumerate() {
         let n = match item {
-            Value::Int(n) => n.to_i64().ok_or_else(|| {
-                RuntimeError::Error(format!("{}: payload must be a List<Int>", method))
-            })?,
+            Value::Int(n) => match n.to_i64() {
+                Some(n) => n,
+                // An `Int` outside `i64` is still an `Int` — a fortiori out
+                // of byte range, so it takes the catchable value-error path,
+                // not the type-error one.
+                None => {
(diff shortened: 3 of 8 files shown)


jasisz (commented):
Reviewed, and ran it locally — both new tests pass (including the integration one with `--include-ignored`), fmt is clean. This is a well-made first PR: the wiring is complete across the VM path, the additive sibling is the right call, and the docs/`llms.txt` steering toward `sendBytes` for binary work is appreciated.

Answers to your open questions:

- **`List<Int>` is right for now.** Consistency with the `Byte` namespace wins at this stage; if parsing pressure shows `Vector<Int>` should be the primary shape, that's a pre-1.0 revisit, not something to solve here.
- **VM-only is fine for this PR.** I'd take the Rust codegen arm as a follow-up rather than grow this one. *(Edited:* I originally echoed the "wasip2 is host territory" line from the PR description here — that's stale, and the mistake is on me since it's my own code: `Tcp.*` graduated on `--target wasip2` in 0.20 with an in-module wasi-sockets implementation, and plain `--target wasm-gc` routes `Tcp.send` through an `aver.tcp_send` host import. So both wasm targets can and will carry `sendBytes` — I'm taking that as follow-up PRs alongside the Rust codegen arm.*)*

One fixup before I merge: in `bytes_arg`, an `Int` outside `i64` falls into the `to_i64()` failure and reports "payload must be a List<Int>". Aver's `Int` is arbitrary-precision, so this is reachable from ordinary source, and it lands as a runtime trap with a wrong message:

```
[65, 256]                        → Result.Err("… byte 256 at index 1 is out of range (0–255)")   — catchable, as designed
[65, 1208925819614629174706176]  → Runtime error: Tcp.sendBytes: payload must be a List<Int>     — trap, and the message is wrong
```

By your own (correct) design both belong in the catchable out-of-range `Result.Err`, index included — treating `to_i64()`'s `None` the same as the `u8::try_from` failure should do it.

The two things you hit while testing deserve their own issues — the Rust codegen reporting `Compiled [Rust]` and then failing at `cargo build`, and the wasm-gc `List<Int>` literal rejection. Would you file them with your repros? You found them, you should get the credit.

CI and Proof are already green here; once the fixup is in and everything stays green, I'll merge.

jasisz:
Update on the two side-findings: the wasm-gc `List<Int>` literal rejection turned out to be a discovery gap in the backend's type registry and is fixed in #766 — no need to file that one. The Rust codegen reporting success and then failing at `cargo build` is still worth an issue if you're up for it.

jasisz:
Didn't want your first PR hanging on a 3-liner, so I pushed the fixup myself (346f3f7) — an `Int` outside `i64` now takes the same catchable out-of-range `Result.Err` as `256`, index included, with a regression test alongside yours. Hope you don't mind. Merging once CI is green — thanks again, this makes binary responses actually reachable from Aver.

PR #767, opened 2026-07-30
Pin the declared function type of every host-table helper

jasisz:
Three conjuncts on the expression-fragment acceptance arm, each closing a channel the byte-equality gate does not read.

## What the branch pins

**1. Every cited host helper's declared function type** (`WasmSlice.hostTableFuncTypesMatch`). Helper bodies are pinned by template byte equality, but a body equality says nothing about the declared type: a helper declared at a strict supertype of the carrier reference still validates by subtyping while the proof faces model the exact claimed carrier. Each `(role, funcIdx)` entry must now declare exactly the signature its role fixes over the claimed carrier (`box`: `[i64] -> [(ref null carrier)]`, `toIndex`: `[(ref null carrier)] -> [i32]`, `add`/`sub`/`mul`: the canonical two-argument carrier signature), resolved through the function section with imports failing closed. Applied at all six acceptance sites.

**2. The carrier reference point itself** (`AcceptedArtifact.symFragmentCarrierBound`). For expression fragments the claimed carrier is producer data no decoder constrains, so pin 1 alone is circular on that family: declare the box helper at a fake supertype, claim that same fake index, and the pin holds while the template-pinned body still builds structs at the real carrier. The claim must now present `CertDecode.carrierState modBytes modLen = some (some carrier)` — a byte-provably carrierless module (`some none`) and an undecodable type section (`none`) both reject.

The trigger is derived from the claim, not from a hand-listed set of families: the binding applies when the host table is non-empty **or** the encoded plan names `intCarrier` anywhere — parameters, result, or any node type in the body or in any nested block. Keying it on table emptiness alone was unsound: `StandardFace.fragment` states `domRepr := args = FragParams.encodeArgs carrier params values`, and `FragTy.encodeArg` sends an `intCarrier` parameter to `carrierSmall carrier value`, a concrete three-field struct at the claimed index. The `Int -> Bool` comparison family asserts that layout while citing no role at all, so a table-keyed exemption covered a carrier-sensitive face. The plan walk lists every `FragNodeKind` constructor explicitly (a new block-carrying node makes the match non-exhaustive and stops the wall building) and answers "required" when out of fuel. The residual permissive case is exactly "no role cited and no `intCarrier` in the plan", where no face can mention the index — that is what keeps projection and float/string-boundary fragments certifiable in modules without a carrier.

**3. The width of a face's asserted value layout** (`WasmSlice.checkTagDispatchTypes`). The tag-dispatch face states `domRepr := vs = [.structv optIdx [.i32v tag, payload]]` — exactly two fields — while the type check required only `structIdx != carrier` and field 0 being the i32 tag. A wider scrutinee left the obligation quantified over states the module's own type section forbids. The decoded field count is now pinned to two, matching the pin `checkExprProjectionTypes` already carried for the projection face. The fused vector-read face is not in this shape: `vecDomRepr` quantifies its element list existentially, so it asserts no width and pins the element type instead.

## Attribution

`tests/cert_whole_module_guard_iso.rs` carries the matrix. Every weakened copy is cut from the LIVE materialized wall source at test time, with exactly-once surgery assertions — never a hand-maintained copy. For each hostile shape the real predicate rejects, the copy weakened by exactly that conjunct accepts, and copies weakened by the others still reject; each honest twin is accepted by the real predicate and by every weakened copy, so no rejection is a framing artefact.

- `hostileSig` — attributed to the declared-type pin.
- `hostileCarrier` — box declared at a fake supertype and the claim citing that same index; attributed to the carrier binding, with the declared-type pin proven to hold of it.
- `hostileBoth` — rejected by each singly-weakened copy and **accepted by the doubly-weakened copy**, which is built rather than asserted.
- `genericHostile` — a role-free `Int -> Bool` fragment claiming a four-field struct; attributed to the carrier binding, with the declared-type pin proven vacuously true on the empty table. On an empty table the previous shape of the conjunct reduced to `True`, so the copy weakened by the whole conjunct is exactly that shape on this input: the acceptance is the hole, exhibited.
- `wideOpt` — the honest tag-dispatch module with a three-field scrutinee; attributed to the arity term through a copy of the whole `checkTagDispatchTypes` -> `exprTagDispatchTypesMatch` -> `exprFragmentNominalTypesMatch` -> `exprFragmentPlanAccepted` -> `symFragmentPlanAccepted` chain, re-cut from the live sources with only that term removed.

The scope of the permissive arm is exhibited too: the walk is shown to descend two `ifElse` levels, and a plan that names no carrier is shown to admit even a nonsense index — which is the whole content of the exemption.

## Format specification

`docs/certificate-format.md` documented only the declared-type pin with normative MUST language, so a reimplementor following it would have built a checker with exactly the hole conjunct 2 closes. Both conjuncts are now documented with the same force, with what each closes and the statement that neither subsumes the other; the face-layout arity pins are documented alongside. Two stale enumerations in the same section (`HostRole` missing `toIndex`, `FragPrim` missing `i32Eq`/`i32And`) are corrected.

## Gates

`cert_certify_spec` 34/34, `cert_whole_module_guard_iso` 12/12, `cert_decode_spec` 7/7, `aver-cert` wall tests 6/6. End-to-end `aver cert verify` CERTIFIED (exit 0) on `tools/certkit/fixtures/rangepred.av` (both triggers: five roles and an `intCarrier` parameter), `examples/core/hello.av` (carrierless module), and `tools/certkit/fixtures/letnamed.av` (tag dispatch, the face whose scrutinee width is newly pinned). Wall builds clean, capstone axioms exactly `[propext, Classical.choice, Quot.sound]`, no `sorry`/`admit`. `cargo fmt --check` and clippy in all three CI shapes clean. Wall id `sha256:03080a8c...` -> `sha256:796f3251...`; the `add_one` snapshot drifts by exactly that line.


jasisz:
Adversarial review found a MAJOR that this branch does not close, so it should not merge as it stands.

The declared-type conjunct is circular on the expression-fragment arm — the one the ticket names as the live one. `symFragmentPlanAccepted` passes the claim's own `carrier` field, which nothing binds to the module bytes (this file says so about itself at AcceptedArtifactCore.lean:1499-1516, and `symFragmentClaims` is deliberately absent from `decodedNonExprClaimFacts`). A wasm declaration is an upper bound on what a body returns, so a producer can declare the box helper at the fake supertype **and** claim that same index: the new conjunct is satisfied, the template body pin still passes (it never reads a declared type), the module validates by subtyping, and the tag-dispatch face still asserts a box the bytes contradict.

The other four families are unaffected — their carrier is byte-derived through `decodedStrictCarrierIndex`, so the pin closes the channel there as intended.

Fix in progress on this branch: a complementary conjunct binding a role-citing fragment's carrier to `CertDecode.carrierState`, with empty host tables left unconstrained so carrierless generic and projection fragments keep certifying. The two conjuncts split the work — the signature pin catches a wrong helper type under a right carrier, the carrier binding catches both being consistently fake — so the isolation test will attribute each separately.

jasisz:
Second review round found more, all confirmed against refutation, so this still does not merge.

**The empty-table exemption is unsound.** The carrier binding added last round is keyed on the host table being non-empty, on the theory that a claim citing no role has no carrier-sensitive face. That theory is wrong: the generic fragment face's domain relation encodes an `intCarrier` parameter as a three-field carrier value at the claimed index, and the generic admission rule forbids `intCarrier` results while permitting `intCarrier` parameters. The Int-versus-constant comparison family consults the host table for nothing, so it reaches acceptance with an empty table and a face asserting a layout the bytes never confirm. Three reviewers found this independently.

**The tag-dispatch face fixes its scrutinee at exactly two fields** and no conjunct binds that arity — the type check requires only that the index differ from the carrier and that field 0 be an i32.

**The spec documents one half of the pin** with normative MUST language and is silent on the other, which would guide a reimplementation into the hole the previous round closed.

Round 3 in progress on this branch: the decoder equality gets keyed on what the face actually models, derived by walking the plan for carrier-typed positions rather than on table emptiness; the scrutinee field count gets pinned; the spec gets both conjuncts. The isolation matrix grows a probe per attack, keeping the exclusive-attribution discipline.

Worth recording: all three rounds found the hole in the same place — an exemption justified by prose rather than derived mechanically. An exemption in a trusted predicate needs a derivation, not an argument about which cases are affected.

PR #777, opened 2026-08-04
Add Tcp.readBytes — byte-clean, exact-length reads on a persistent connection

### LINKED ISSUE #776: `Tcp.readLine` cannot read binary frames, and there is no exact-length read
Split out of #753 (Issue 2 of the four described there) so that a fix can close
this one on its own. Sibling of #763, fixed by #764 — that covered the one-shot
path; this is the persistent-connection path.

## What happens

`Tcp.readLine` is the only way to read from a `Tcp.Connection`, and it fails on
binary data in two independent ways.

`aver-rt/src/tcp.rs:80` reads through `BufRead::read_line`, then strips a
trailing `\r\n`.

### 1. Non-UTF-8 input is a hard error

`BufRead::read_line` requires valid UTF-8 and returns `ErrorKind::InvalidData`
otherwise. Any byte sequence that isn't valid UTF-8 makes the read fail
outright:

```
Result.Err("stream did not contain valid UTF-8")
```

Note this is the *opposite* symptom to #763. There, `Tcp.send` silently
corrupted the data via `from_utf8_lossy`; here the read refuses. Same underlying
assumption — that a socket carries text — but the two paths fail differently,
which makes the shared cause harder to see from the Aver side.

### 2. Framing splits payloads at `0x0A`

Even when the data *is* valid UTF-8, `read_line` frames on `\n`. A
length-prefixed protocol carries `0x0A` at arbitrary offsets inside payloads,
so messages are cut at meaningless boundaries. Given a 5-byte payload
`41 42 0A 43 44`:

```
server sent 5 bytes: 41 42 0A 43 44
readLine returned: Result.Ok("AB")
```

Two bytes returned, one silently consumed as a delimiter, two left in the
buffer. The caller has no way to know this happened. The trailing-`\r\n` strip
compounds it: those bytes are destroyed when they're payload rather than
framing.

### 3. There is no way to read a fixed number of bytes

This is the missing capability behind both. Length-prefixed framing is
universally "read a fixed-size header, decode a length field, read exactly that
many bytes." No combination of the current `Tcp` API expresses an exact-length
read.

## What I expected

A way to read a caller-specified number of bytes off a persistent connection
without the runtime interpreting, reframing, or decoding them.

## Steps to reproduce

1. Start a server that writes binary and holds the connection open:

   ```python
   import socket, threading, time
   srv.bind(("127.0.0.1", 18401)); srv.listen(5)
   def handle(c):
       c.sendall(bytes([0xF9, 0xBE, 0xB4, 0xD9]))   # case 1: non-UTF-8
       # c.sendall(b"AB\nCD")                        # case 2: valid UTF-8, embedded 0x0A
       time.sleep(3); c.close()
   while True:
   ```

2. Read from it:

   ```aver
   module ReadLineRepro
       intent = "Read a binary frame over a persistent connection."
       exposes [main, readIt]
       effects [Tcp.connect, Tcp.readLine, Console.print]

   fn readIt() -> Result<String, String>
       ? "Open the connection and read one line."
       ! [Tcp.connect, Tcp.readLine]
       conn = Tcp.connect("127.0.0.1", 18401)?
       Tcp.readLine(conn)

   fn main() -> Unit
       ? "Show what readLine returns."
       ! [Tcp.connect, Tcp.readLine, Console.print]
       Console.print("{readIt()}")
   ```

3. Case 1 gives `Result.Err("stream did not contain valid UTF-8")`; case 2 gives
   `Result.Ok("AB")` from a 5-byte payload.

## Proposed fix

An additive byte-clean, exact-length read, mirroring how #764 added
`Tcp.sendBytes` beside `Tcp.send` rather than reshaping it:

```
Tcp.readBytes : (Tcp.Connection, Int) -> Result<List<Int>, String>
```

Semantics I'd suggest, all open to a different call:

- **Exact length.** Reads exactly `n` bytes. A short read is an error, not a
  truncated success — for length-prefixed framing, fewer bytes than the prefix
  promised means the peer went away mid-message, and silently returning a
  partial frame would desynchronise the caller's parser.
- **Capped.** `n` bounded by the existing `BODY_LIMIT` (10 MiB). Length
  prefixes arrive from an untrusted peer — Bitcoin's is 4 bytes, so it can ask
  for 4 GiB — and the cap keeps a hostile or corrupt prefix from becoming an
  allocation of that size.
- **Catchable errors.** A negative or oversized `n` returns `Result.Err` rather
  than trapping, consistent with the port range and with the bignum handling
  you added while merging #764.

`Tcp.readLine` stays exactly as it is for text protocols.

## Related

`Tcp.writeLine` has the mirror-image problem on the write side (Issue 3 of
#753): it appends `\r\n` unconditionally and its `String` parameter is UTF-8,
so `0xF9` cannot be expressed. `readBytes` and a corresponding `writeBytes` are
both needed before a persistent binary session is possible — this issue covers
only the read half.

n1bor:
# Add `Tcp.readBytes` — byte-clean, exact-length reads on a persistent connection

Fixes #776. Companion to #764 (`Tcp.sendBytes`): that covered the one-shot path,
this is the session path. Broader context in #753.

## The problem

`Tcp.readLine` is the only way to read from a `Tcp.Connection`, and it fails on
binary data twice over. Both demonstrated against a local server:

**Non-UTF-8 is a hard error.** `BufRead::read_line` requires valid UTF-8:

```
server sent: F9 BE B4 D9
readLine   : Result.Err("stream did not contain valid UTF-8")
```

**Valid UTF-8 gets silently reframed.** `read_line` frames on `\n`, which
appears at arbitrary offsets inside binary payloads:

```
server sent: 41 42 0A 43 44   (5 bytes)
readLine   : Result.Ok("AB")   (2 bytes, one consumed as a delimiter)
```

The second is the more dangerous of the two — it succeeds with the wrong answer.
The trailing-`\r\n` strip compounds it when those bytes are payload.

Underneath both: there is no way to read a fixed number of bytes, which is what
every length-prefixed protocol needs.

## The change

Adds `Tcp.readBytes : (Tcp.Connection, Int) -> Result<List<Int>, String>`.
Reads exactly `n` bytes off the connection and decodes nothing. `Tcp.readLine`
is untouched and stays the right tool for line-oriented text.

Reading a real length-prefixed frame — 4-byte magic, 4-byte LE length, then the
payload it announces, containing two `0x0A`s and a `0xFF`:

```
server sent: [249,190,180,217, 5,0,0,0, 1,10,255,10,3]
readBytes:   Result.Ok([249, 190, 180, 217, 5, 0, 0, 0, 1, 10, 255, 10, 3])
```

### Design decisions

**Short read is an error, not a truncated success.** Fewer bytes than the
length prefix promised means the peer went away mid-message. Returning a partial
frame would desynchronise the caller's parser at a point far from the cause.

**Capped at `BODY_LIMIT` (10 MiB).** Length prefixes come off the wire from an
untrusted peer — Bitcoin's is four bytes, so it can ask for 4 GiB. Without a cap
a hostile or corrupt prefix becomes an allocation of that size. This is the one
decision I'd most like a second opinion on: the cap matches `Tcp.send`, but a
caller wanting to stream something larger currently has no way to raise it.

**Count errors are catchable, never traps.** Negative, over-cap, and
`i64`-overflowing counts all return `Result.Err`. The bignum case follows the
correction you made when merging #764 — an `Int` too large for `i64` is still an
`Int`, so it takes the value-error path rather than being reported as a bogus
type error.

**`List<Int>` carrier** for symmetry with `Tcp.sendBytes`.

## What's wired

- `aver-rt::tcp::read_bytes`
- VM path: dispatch, `register`/`register_nv`, `DECLARED_EFFECTS`, `effects`,
  `VmBuiltin`
- Effect classification: `GenerativeOutput`, matching the other session methods
- **Hostile profiles** — `normal_ok`, `short_read`, `always_err`. I omitted
  these for `Tcp.sendBytes` and you had to add them in #772; the completeness
  guard from that PR passes here
- Docs: `services.md`, `oracle.md` (both tables), `llms.txt`

## What's not wired

Codegen backends, same as #764 — this is VM-only. `wasip2.md` is deliberately
not updated, since I haven't implemented the wasi-sockets path. Happy to extend
if you'd point me at the shape you want.

## Tests

Three added to `tests/eval_spec.rs`:

- `tcp_read_bytes_round_trips_binary_frame` — header + exact-length payload
  containing `0x0A` and `0xFF`, the bytes that break `readLine` both ways
- `tcp_read_bytes_short_read_is_an_error` — peer sends 3 bytes, 10 requested
- `tcp_read_bytes_rejects_out_of_range_counts` — negative, over-cap, and bignum

Full suite green, `cargo fmt --all -- --check` clean.

## One thing I noticed — and a correction

`AGENTS.md:39` says `Tcp.Connection` is "No longer opaque — constructable via
`Tcp.Connection(id = ..., host = ..., port = ...)`". That is stale.
`docs/services.md` is the accurate one: construction is rejected, and
`aver check` says so plainly —

```
error[type-error]: Cannot construct opaque type 'Tcp.Connection' — use its module's constructor function
```

An earlier version of this description had those two the wrong way round. My
apologies — I'd checked `AGENTS.md` and the test suite but not `aver check`
itself.

It mattered here. `tcp_read_bytes_rejects_out_of_range_counts` originally
constructed a `Tcp.Connection` so the count checks could run without a socket.
It passed only because `call_fn_with_effects` in `tests/eval_spec.rs` parses and
VM-compiles without running the type checker, so the opaque rule never fires —
i.e. the test exercised a construction real Aver source cannot perform. It now
uses a live listener instead.

Worth flagging separately: that harness gap means any test written through
`call_fn_with_effects` can silently depend on programs the type checker would
reject. Happy to open an issue if that's useful.


jasisz:
Maintainer follow-up (2026-08-06): I extended this PR without replacing the original implementation or description.

The final public signature is `Tcp.readBytes : (Tcp.Connection, Int) -> Result<Bytes, String>`, using the nominal `Bytes` type shared with `Tcp.sendBytes`; callers can inspect it through `Bytes.toList`.

This branch now covers:

- the original VM implementation and hostile profiles,
- generated Rust,
- hosted wasm-gc, including effect recording and replay,
- native WASI 0.2 sockets,
- the wasm-gc certificate capability registry and current wall identity.

Added cross-backend coverage includes exact binary reads split across multiple socket chunks, non-UTF-8 payloads, short-read errors, arbitrary-precision/out-of-range counts as catchable `Result.Err`, generated-code compilation, wasm validation, and hosted record/replay.

Self-host remains deliberately outside this PR, as previously agreed.

PR #779, opened 2026-08-04
Add Tcp.writeBytes — put exact bytes on a persistent connection

### LINKED ISSUE #778: `Tcp.writeLine` cannot put arbitrary bytes on the wire
Split out of #753 (Issue 3 of the four described there) so that a fix can close
this one on its own. Completes the pair with #776 (`Tcp.readBytes`, read side)
and #763 (`Tcp.sendBytes`, one-shot path, fixed by #764).

## What happens

`Tcp.writeLine` is the only way to write to a `Tcp.Connection`, and it cannot
express an arbitrary byte sequence. Two independent reasons, both visible on the
wire.

`aver-rt/src/tcp.rs:63`:

```rust
let msg = format!("{}\r\n", line);
reader.get_mut().write_all(msg.as_bytes())
```

### 1. `\r\n` is appended unconditionally

Every write gets two extra bytes. For a length-prefixed protocol this
desynchronises the stream immediately — the peer reads the trailing `\r\n` as
the beginning of the next message, and every subsequent frame is misaligned.

### 2. Bytes above `0x7F` cannot be expressed at all

The parameter is an Aver `String`, which maps to a Rust `String` and is
therefore always valid UTF-8. `as_bytes()` UTF-8-encodes it, so a codepoint
above `0x7F` goes out as a multi-byte sequence rather than the single byte the
protocol requires.

Measured against a server that logs exactly what it receives:

```
Aver sent "AB"          →  wire: 41 42 0D 0A
Aver sent "ù" (U+00F9)  →  wire: C3 B9 0D 0A
```

The second line is the crux. To put the byte `F9` on the wire there is no
spelling that works: `"ù"` becomes `C3 B9`, and Aver has no byte-string literal
or `String.fromBytes` escape hatch. Both calls return `Result.Ok`, so nothing
signals that the bytes sent were not the bytes intended.

For Bitcoin specifically, mainnet framing begins `F9 BE B4 D9` — so the
handshake cannot even be started.

## What I expected

A way to write a caller-supplied byte sequence to a persistent connection
exactly as given, with nothing appended and nothing re-encoded.

## Steps to reproduce

1. Start a server that logs the bytes it receives:

   ```python
   import socket, threading
   srv.bind(("127.0.0.1", 18600)); srv.listen(5)
   def handle(c):
       c.settimeout(2.0); data = b""
       try:
           while True:
               chunk = c.recv(4096)
               if not chunk: break
               data += chunk
       except Exception: pass
       print("RECEIVED:", " ".join(f"{b:02X}" for b in data), flush=True)
       c.close()
   while True:
   ```

2. Write to it:

   ```aver
   module WriteLineRepro
       intent = "Show what writeLine actually puts on the wire."
       exposes [main, sendIt]
       effects [Tcp.connect, Tcp.writeLine, Tcp.close, Console.print]

   fn sendIt(payload: String) -> Result<Unit, String>
       ? "Write one payload and close."
       ! [Tcp.connect, Tcp.writeLine, Tcp.close]
       conn = Tcp.connect("127.0.0.1", 18600)?
       _w = Tcp.writeLine(conn, payload)?
       Tcp.close(conn)

   fn main() -> Unit
       ? "Two cases: plain ASCII, and a codepoint above 0x7F."
       ! [Tcp.connect, Tcp.writeLine, Tcp.close, Console.print]
       Console.print("{sendIt("AB")}")
       Console.print("{sendIt("ù")}")
   ```

3. The server logs `41 42 0D 0A` and `C3 B9 0D 0A`. Both Aver calls returned
   `Result.Ok`.

## Proposed fix

An additive byte-clean write, mirroring `Tcp.sendBytes` (#764) and the proposed
`Tcp.readBytes` (#776):

```
Tcp.writeBytes : (Tcp.Connection, List<Int>) -> Result<Unit, String>
```

Writes the bytes exactly as given: nothing appended, nothing encoded. Suggested
details, all open to a different call:

- Elements outside `0..=255` return a catchable `Result.Err` naming the value
  and its index — the same shape `Tcp.sendBytes` uses, including the bignum
  handling added while merging #764.
- An empty payload is a no-op returning `Result.Ok(Unit)`.
- `Tcp.writeLine` stays exactly as it is for line-oriented text protocols.

With this and #776, a persistent binary session becomes expressible: connect,
write a framed request, read a fixed-size header, read exactly the payload it
announces.

## Related

This is the last of the three TCP blockers in #753. Issue 4 there (no SHA-256
callable from Aver) is independent and still needed for protocols that checksum
their frames.

n1bor:
# Add `Tcp.writeBytes` — put exact bytes on a persistent connection

Fixes #778. Completes the set with #764 (`Tcp.sendBytes`, one-shot, merged) and
#777 (`Tcp.readBytes`, session read). Parent: #753.

## The problem

`Tcp.writeLine` is the only way to write to a `Tcp.Connection`, and it cannot
express an arbitrary byte sequence. Measured against a server that logs exactly
what it receives:

```
Aver sent "AB"          →  wire: 41 42 0D 0A
Aver sent "ù" (U+00F9)  →  wire: C3 B9 0D 0A
```

Two separate defects in `aver-rt/src/tcp.rs:63`:

```rust
let msg = format!("{}\r\n", line);
reader.get_mut().write_all(msg.as_bytes())
```

**`\r\n` is appended unconditionally.** Two bytes nobody asked for. In a
length-prefixed protocol the peer reads them as the start of the next message
and every subsequent frame is misaligned.

**Bytes above `0x7F` cannot be expressed.** The parameter is an Aver `String`,
so it is valid UTF-8 by construction and `as_bytes()` re-encodes: U+00F9 goes
out as `C3 B9`, not `F9`. There is no byte-string literal and no
`String.fromBytes` to reach for instead. The byte simply cannot be sent.

Both calls returned `Result.Ok`. Nothing tells the caller the wrong bytes went
out.

## The change

Adds `Tcp.writeBytes : (Tcp.Connection, List<Int>) -> Result<Unit, String>`,
writing the payload exactly as given. `Tcp.writeLine` is untouched and remains
right for line-oriented text.

Same server, after:

```
intended: F9 BE B4 D9 05 00 00 00 01 0A FF
wire:     F9 BE B4 D9 05 00 00 00 01 0A FF
```

`F9` present, `0x0A` carried through the middle of the payload rather than
treated as a delimiter, and nothing appended.

### Design decisions

**Validation before any wire I/O.** An out-of-range element is rejected before a
single byte is written, so a bad payload cannot half-write and leave the peer
holding a truncated frame. There is a test asserting the server receives nothing
in that case.

**Errors are catchable and reuse `bytes_arg`.** Elements outside `0..=255`
return `Result.Err` naming the value and its index. Sharing the helper with
`Tcp.sendBytes` means the bignum case is handled the way you settled it when
merging #764 — an `Int` too large for `i64` is still an `Int`, so it takes the
value-error path rather than a bogus type error.

**Empty payload is a no-op** returning `Result.Ok(Unit)`, matching `write_all`
on an empty slice.

**`List<Int>` carrier** for symmetry with `sendBytes` and `readBytes`. Writes
are the one place `Vector<Int>` would buy nothing, since the payload is
serialised to `Vec<u8>` immediately either way — so consistency across the three
byte APIs seemed worth more than the alternative. Happy to be told otherwise;
it is a one-line change while there are no callers.

## What's wired

- `aver-rt::tcp::write_bytes`
- VM path: dispatch, `register`/`register_nv`, `DECLARED_EFFECTS`, `effects`,
  `VmBuiltin`
- Effect classification: `GenerativeOutput`, matching the other session methods
- Hostile profiles (`normal_ok`, `always_err`), so the completeness guard from
  #772 stays satisfied
- Docs: `services.md`, `oracle.md`, `llms.txt`

## What's not wired

Codegen backends — VM-only, same scope as #764 and #777. `wasip2.md` is
deliberately not updated since the wasi-sockets path isn't implemented here.

## Tests

Two added to `tests/eval_spec.rs`:

- `tcp_write_bytes_puts_exact_bytes_on_the_wire` — the server thread captures
  the raw bytes and the test asserts on them directly, which is the only way to
  catch an unwanted `\r\n` or a re-encoded byte
- `tcp_write_bytes_rejects_out_of_range_byte` — asserts both the error message
  and that the server received nothing

Both use a real listener rather than a constructed `Tcp.Connection`: that type
is opaque and the type checker rejects construction, as noted on #777.

Full suite green (89 suites, 2797 passed), `cargo fmt --all -- --check` clean,
and all three CI clippy invocations pass with `-D warnings`.

## Together with #777

These two are independent in code — different functions, no shared lines — so
they can land in either order. Together they make a persistent binary session
expressible for the first time: connect, write a framed request, read a
fixed-size header, read exactly the payload it announces.


jasisz:
Maintainer follow-up (2026-08-06): I extended this PR on top of the now-merged #777 without replacing the original implementation or description.

The final public signature is `Tcp.writeBytes : (Tcp.Connection, Bytes) -> Result<Unit, String>`. Validation belongs at the nominal refinement boundary (`Bytes.fromList` / `Bytes.fromHex`) before any socket I/O; the runtime bridges also fail closed if a malformed internal carrier reaches them.

This branch now covers:

- the original VM implementation and hostile profiles,
- generated Rust,
- hosted wasm-gc, including effect recording and replay,
- native WASI 0.2 persistent sockets,
- the wasm-gc certificate capability registry and current wall identity.

Cross-backend tests cover exact non-UTF-8 bytes with an embedded newline and verify that no text encoding or CRLF framing is added. The native WASI test additionally exposed and fixed an existing helper-allocation bug: `writeBytes + readLine` could accidentally allocate the `readBytes` helper based on coincidental types/imports; byte helpers are now gated by their actual declared effects.

Local gates are green: 1007 library tests, both clippy passes, 39 `aver-cert` tests, VM live sockets, hosted wasm-gc record/replay, generated-Rust compilation, wasm validation, and a native WASI live socket probe.

Self-host remains deliberately outside this PR, as previously agreed.

PR #781, opened 2026-08-04
Add a pure Crypto namespace with Crypto.sha256

### LINKED ISSUE #780: No SHA-256 callable from Aver, and no way to implement one in the language
Split out of #753 (Issue 4 of the four described there) so that a fix can close
this one on its own. The other three were the TCP layer — #763 (fixed by #764),
#776 (PR #777), #778 (PR #779). This one is independent of all of them: even
with byte-clean sockets, a binary protocol that checksums its frames still can't
be spoken.

## What happens

There is no hashing surface in the language. Nothing in the `Crypto` namespace,
because there is no `Crypto` namespace:

```
error[unknown-ident]: Unknown identifier 'Crypto'
  at: nohash.av:6:1
  repair: Check the spelling or add the missing import
```

`docs/services.md` lists the full namespace set — Bool, List, Vector, Result,
Option, Int, Float, String, Map, Char, Byte, Args, Console, Http, HttpServer,
Disk, Tcp, Random, Time, Terminal, Env — and no hashing appears anywhere in it.

## Why it can't be worked around in Aver

SHA-256 is defined over 32-bit words in terms of XOR, AND, NOT, right-rotation
and right-shift. Aver has no bitwise operators:

```
error[parse-error]: Unknown character: '^'
  at: noxor.av:8:8
```

This holds in the implementation, not just the docs — no bitwise tokens exist in
the lexer or parser:

```
grep -rniE "caret|ampersand|shiftleft|shiftright|bitand|bitor|xor" src/lexer.rs src/parser/*.rs
→ no matches
```

Simulating them arithmetically through `Int.div` / `Int.mod` is possible in
principle and impractical in fact: both return `Result<Int, String>`, so every
one of roughly sixty operations per round needs unwrapping; `Int` is `i64` with
no unsigned 32-bit wrapping type; and SHA-256 runs 64 rounds per 512-bit block.
The result would be large, slow, and unreadable — the opposite of what Aver
optimises for.

So this is not a case where a library or a user-space implementation could fill
the gap. It has to come from the runtime.

## What I expected

To be able to hash a byte sequence. Concretely, the Bitcoin message checksum is
the first four bytes of a double-SHA-256 over the payload, and every transaction
id is a double-SHA-256 of the serialised transaction. Neither is expressible.

## Steps to reproduce

```aver
module NoHash
    intent = "Try to compute a SHA-256 checksum."
    exposes [checksum]
    effects []

fn checksum(payload: List<Int>) -> List<Int>
    ? "Bitcoin message checksum: first 4 bytes of double-SHA256."
    Crypto.sha256(Crypto.sha256(payload))
```

`aver check` reports `Unknown identifier 'Crypto'`.

## The implementation is already in the tree

This is the part that makes the request cheap. `sha2` is already a dependency of
`aver-lang`, declared optional and switched on by the `wasm-compile` feature:

```toml
# Cargo.toml
sha2 = { version = "0.10", optional = true }
wasm-compile = ["dep:aver-cert", "dep:wasm-encoder", "dep:wasmprinter", "dep:wasmparser", "dep:wat", "dep:sha2"]
```

It is already resolved in `Cargo.lock` (`sha2 0.10.9`) and already linked into
any binary built with `--features wasm`. The gap is that it exists only as a
compile-time implementation detail of artifact certification, with no surface
exposed to user programs.

## Proposed fix

A pure `Crypto` namespace, following the `Byte` pattern in `src/types/byte.rs` —
a namespace of operations over existing types rather than a type of its own:

```
Crypto.sha256 : List<Int> -> Result<List<Int>, String>
```

Suggested details, all open to a different call:

- **Pure, not effectful.** Hashing is deterministic and total given valid input,
  so it needs no `! [...]` declaration. That matters beyond tidiness: it means
  hashing code can carry ordinary `verify` blocks with the published NIST test
  vectors, rather than needing Oracle stubs.
- **`Result` return** for symmetry with the byte APIs — elements outside
  `0..=255` return a catchable `Result.Err` naming the value and its index, the
  same shape `Tcp.sendBytes` uses.
- **`List<Int>` in and out**, 32 elements out, consistent with the `Tcp` byte
  methods.
- **Dependency promotion.** `sha2` moves out of `wasm-compile` so `Crypto` is
  available in the default feature set. A hashing builtin that only exists in
  `--features wasm` builds would be a confusing surface.

SHA-512 is nearly free from the same crate if it is wanted. RIPEMD-160 would be
a genuinely new dependency and is only needed for address derivation, so it can
reasonably come later or never — I have deliberately left it out.

## Scope note

I would expect this to be VM-only in a first pass, like #764/#777/#779, with the
codegen backends following separately. Happy to be told otherwise.

n1bor:
# Add a pure `Crypto` namespace with `Crypto.sha256`

Fixes #780. Last of the four in #753; the other three were the TCP layer
(#764 merged, #777 and #779 open). This one is independent of all of them.

## The problem

There is no hashing surface in the language, and no way to build one in it.

```
error[unknown-ident]: Unknown identifier 'Crypto'
```

SHA-256 is defined over 32-bit words in terms of XOR, AND, NOT, rotation and
shift, and Aver has no bitwise operators:

```
error[parse-error]: Unknown character: '^'
```

So this could not be filled by a library or by user code — it had to come from
the runtime. Meanwhile `sha2` was already a dependency, resolved in
`Cargo.lock`, reachable only from the certification pipeline behind
`wasm-compile`.

## The change

```
Crypto.sha256 : List<Int> -> Result<List<Int>, String>
```

A pure namespace following the `Byte` pattern in `src/types/byte.rs` — a
namespace of operations over existing types rather than a type of its own.

Verified against the published FIPS 180-4 vectors, written as an ordinary Aver
`verify` block:

```aver
verify hashOf
    hex(hashOf([]))           => "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
    hex(hashOf([97, 98, 99])) => "ba7816bf8f01cfea414140de5dae2223b00361a396177a9cb410ff61f20015ad"

verify doubleSha
    hex(doubleSha([97, 98, 99])) => "4f8b42c22dd3729b519ba6f68d2da7cc5b2d606d05daed5ad5128cc03e6c6358"
```

All three cross-checked against Python's `hashlib`, and the end-to-end use case
works — a Bitcoin message checksum over `"version"` computes `6fb01d3b` in Aver,
matching `hashlib` exactly.

### Design decisions

**Pure, not effectful.** Hashing is deterministic and total given valid input,
so it carries no `! [...]` and pure callers stay pure. This is the decision I
care most about: it is what makes the verify blocks above possible. Classifying
it as an effect would have forced Oracle stubs onto every hashing test, and
stubbing a hash function means testing against a fiction rather than against the
NIST vectors.

**`Result` return rather than a trap.** Elements outside `0..=255` return a
catchable `Result.Err` naming the value and its index — the same shape
`Tcp.sendBytes` uses, including the bignum rule you settled in #764. It costs a
`?` at the call site, and double hashing composes cleanly through it.

**`List<Int>` in and out**, 32 elements out, consistent with the `Tcp` byte
methods.

**`sha2` promoted out of `wasm-compile`** to a plain dependency, so `Crypto` is
in the default feature set. A hashing builtin that existed only in
`--features wasm` builds would be a confusing surface. `Cargo.lock` is
unchanged — the crate was already resolved.

**Scope.** SHA-512 is nearly free from the same crate if wanted. RIPEMD-160 is
deliberately left out: it is a genuinely new dependency and is only needed for
address derivation.

## One thing worth knowing for the next new namespace

`src/ir/calls.rs` has a hardcoded `is_builtin_namespace` allow-list, and a
namespace missing from it type-checks fine but then fails at
HIR→MIR lowering with:

```
internal error: fn `main` did not lower to MIR (an unsupported shape reached the VM backend)
```

That took a while to track down, because `aver check` reports `✓ types` and the
failure surfaces only at run time, phrased as an internal error rather than as
"this namespace isn't registered". Adding a *method* to an existing namespace
never touches it, so it is easy to miss. Might be worth either a comment
pointing at the list or a clearer diagnostic — happy to open a separate issue if
you think it's worth one.

## What's wired

- `src/types/crypto.rs` — the namespace, `register`/`register_nv`, `call`,
  `call_nv`
- `src/types/mod.rs`, `src/vm/builtin.rs` (enum + dispatch), `src/ir/calls.rs`
- Type checker: signature registration, no effects
- Docs: `services.md`, `llms.txt`, `AGENTS.md` architecture map

## What's not wired

Codegen backends — VM-only, same scope as #764/#777/#779.

## Tests

Four added to `tests/eval_spec.rs`:

- `crypto_sha256_matches_published_vectors` — NIST FIPS 180-4 values, written as
  published rather than captured from this implementation, so a wrong digest
  fails rather than merely a changed one
- `crypto_sha256_composes_for_double_hashing` — the Bitcoin shape
- `crypto_sha256_returns_32_bytes`
- `crypto_sha256_rejects_out_of_range_bytes` — including the bignum case

Full suite green, `cargo fmt --all -- --check` clean, and all three CI clippy
invocations pass with `-D warnings`.


jasisz:
Maintainer update after #788:

The contributor's SHA-256 implementation is preserved, while the branch is now rebased semantically onto Aver's nominal byte types:

```aver
Crypto.sha256 : Bytes -> Digest32
```

Callers import them with `depends [Bytes, Crypto.Digest32]`. SHA-256 remains pure and no longer returns `Result`: `Bytes` already guarantees octets, while `Digest32` guarantees the result length.

The update also:

- projects the opaque `Bytes` carrier at the VM host boundary and constructs a nominal `Digest32`;
- re-canonicalises source-defined types in builtin signatures after embedded modules enter the symbol table;
- removes the obsolete `Byte` merge conflicts from the pre-#788 branch;
- moves the runtime vectors into the embedded stdlib fixture, including empty input, `abc`, double-SHA256, and invalid byte construction.

Local validation: 587 tests passed across the typechecker, VM, eval and stdlib suites; all three CI Clippy configurations pass with `-D warnings`.

The original VM-only backend scope is unchanged. Native Rust, wasm-gc and wasip2 SHA-256 lowering remain follow-up work.

jasisz:
Maintainer follow-up after rebasing on #788:

This PR now carries Crypto.sha256 through every maintained backend except self-host:

- VM/interpreter uses the native SHA-256 implementation.
- Generated Rust uses the shared aver-rt crypto helper.
- wasm-gc and wasip2 use the same internal, host-independent wasm implementation.
- Lean and Dafny receive executable SHA-256 models; there is no Oracle, axiom, assume, or sorry involved.

Validation completed locally:

- 13/13 embedded Bytes, Digest32, and Crypto verify cases pass on VM and wasm-gc, including empty input, abc, the 56-byte padding boundary, invalid Bytes, and double SHA-256.
- Generated native Rust passes cargo check.
- wasip2 produces a component accepted by wasm-tools validate.
- Generated Lean builds with zero sorries.
- Generated Dafny verifies with zero errors and zero axioms, and its executable model passes the FIPS vectors.
- All 1004 aver-lang library tests pass; fmt and clippy are clean.

Crypto remains always available, with no Cargo feature. The source API is Crypto.sha256(Bytes) -> Digest32, where the nominal Digest32 type is owned by the Crypto.Digest32 module.

This is a new maintainer note; the original contributor context above is left unchanged.

PR #787, opened 2026-08-05
Escape Lean reserved identifiers (#783)

### LINKED ISSUE #783: Lean emitter's reserved-word list is incomplete — nine identifiers produce Lean that does not parse
**What happens**

A user function whose name collides with a Lean 4 keyword is emitted verbatim, so the generated proof project fails to parse. `aver check` and `aver verify` pass; only the proof export breaks.

```
error: Bytes.lean:35:3: unexpected token 'at'; expected identifier
error: Bytes.lean:81:9: unexpected token 'at'; expected term
```

**Repro**

```aver
module KwAt
    intent =
        "A user fn named after a Lean keyword."
    exposes [at]
    effects []

fn at(a: Int) -> Int
    ? "Identity, so the only thing under test is the name."
    a

verify at
    at(1) => 1
```

```
aver proof kw.av -o out --check
```

**Confirmed names**

One file per candidate, same shape as above. Names already on the list (`calc`) and non-keywords (`this`, `obtain`) produce zero errors, so the probe discriminates.

| name | parse errors |
|---|---|
| `at` | 2 |
| `using` | 2 |
| `exists` | 2 |
| `sorry` | 2 |
| `suffices` | 2 |
| `variable` | 3 |
| `universe` | 3 |
| `notation` | 3 |
| `attribute` | 3 |

**Where**

`LEAN_RESERVED` in `src/codegen/lean/expr.rs:660`, consumed by `aver_name_to_lean` at `:737`.

**Note on the fix**

Adding nine strings closes the observed cases but not the class — the list is hand-maintained with nothing tying it to the token set of the Lean version in `lean-toolchain`. A guard that fails when an emitted identifier is a keyword would catch the next one.

jasisz:
Closes #783.

This PR:
- escapes identifier-shaped tokens registered by the pinned Lean parser, including all nine reported regressions;
- guards the snapshot against the actual Lean token table so a future toolchain upgrade fails loudly;
- adds an end-to-end Aver-to-Lean fixture checked with lake build;
- repairs the identity_guardrails CI regression currently making main red after 9829cb99.

Local verification:
- cargo fmt --all -- --check
- cargo clippy --lib -- -D warnings
- cargo test --lib codegen::lean:: (128 passed)
- cargo test --test proof_spec export_structure:: (12 passed)
- cargo test --test identity_guardrails (2 passed)

### DIFF
--- a/src/codegen/lean/decl_order.rs
+++ b/src/codegen/lean/decl_order.rs
@@ -164,6 +164,9 @@ fn collect_annotation_type_refs(annotation: &str, out: &mut HashSet<String>) {
 fn collect_type_refs(ty: &crate::types::Type, out: &mut HashSet<String>) {
     use crate::types::Type;
     match ty {
+        // syntax-discovery-only: this per-scope declaration DAG compares the
+        // source spelling against type declarations in the same scope; typed
+        // identity has already selected that scope before this local walk.
         Type::Named { name, .. } => {
             out.insert(name.clone());
         }
--- a/src/codegen/lean/expr.rs
+++ b/src/codegen/lean/expr.rs
@@ -2,6 +2,7 @@
 use super::builtins;
 use super::pattern::emit_pattern;
 use super::shared::to_lower_first;
+pub(crate) use super::syntax::{aver_name_to_lean, lean_name_to_aver};
 use crate::ast::{BinOp, Literal, Spanned};
 use crate::codegen::CodegenContext;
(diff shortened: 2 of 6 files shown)


jasisz:
Pushed 7d81b9f8 to fix the failed generic Check & Test job: proof structure regressions now always inspect generated output, run lake/dafny verification only when those tools are installed, and still pass with both toolchains present. This matches the rest of proof_spec and avoids requiring lake in the generic cargo test job.

PR #788, opened 2026-08-05
Ship Bytes and Crypto.Digest32; retire Byte (#786)

### LINKED ISSUE #786: `List<Int>` cannot express a sequence of bytes
@n1bor — this affects your open PRs, so the apology comes first.

## I was wrong about the signature I asked for

In #753 I asked for `Crypto.sha256 : List<Int> -> List<Int>`, deterministic, no effect declaration. You shipped `Result<List<Int>, String>` and explained why: given an arbitrary `List<Int>`, something has to happen when an element is 300, and a catchable `Result` beats a trap or a silent modulo.

**You were right and I was not.** There were two halves to what I asked for. The totality was fine: hashing can't fail, so it shouldn't carry an error, but the domain was wrong... `List<Int> -> List<Int>` is not a signature that can exist, because something has to happen when an element is 300. I asked for a total function over a type that admits non-octets. Your `Result` is the type system pointing at the wrong domain, and I told you to ignore it.

The same gap sits in all four byte signatures, not just the hash.

## What we found

**The range check exists once per backend, per builtin.** For the merged `Tcp.sendBytes` it lives in the VM, the Rust runtime, the wasm-gc bridge and WASI-p2, where it is assembled byte by byte in emitted WebAssembly with a hand-rolled decimal formatter for the offending value. Three separate test files pin the message so the copies stay in step. Your `Crypto.sha256` PR (#781) adds another copy of the helper, correctly, because it sits in a different namespace. That cost repeats for every byte-carrying builtin, and three more are in flight.

**The type lies about the result too, not just the argument.** SHA-256 returns exactly 32 octets and cannot fail. The type says neither. Length matters here: your checksum takes the first four bytes of the digest, which only works because there are at least four.

**The obvious workaround doesn't work.** I wrote the byte type as an ordinary Aver module: opaque record, validating smart constructor, the pattern the docs advertise. I expected to close this without touching the compiler. It runs, it verifies, its laws prove. But the invariant never reaches the prover. A scalar carrier lifts to a Lean subtype; a container carrier is silently emitted as a plain record. So every law still restates "all elements are octets" by hand, which is what the type was supposed to remove. Filed separately as #783, #784 and #785, the silent one being the worst.

## Options

**A. Keep `List<Int>` as the public type.** No new machinery. The per-backend validator multiplies with every new byte builtin, `Result` keeps mixing "the socket failed" with "your element was 300", and every law about bytes carries a premise someone has to remember.

**B. A nominal scalar `Byte`, sequences stay `List<Byte>`.** The whole first-order `List` API keeps working. Costs a second scalar numeric type in a language that deliberately has one mathematical `Int`, and parsing code reads `Byte.toInt(b) * 256 + Byte.toInt(next)` throughout.

**C. `Bytes` as a compiler builtin.** Solves bytes, quickly and cleanly. Solves only bytes. A 32-byte digest, a 4-byte magic, a txid, a var-int are each another compiler change, and you get none of them yourself.

**D. Make the existing refinement mechanism general, with `Bytes` as its first client.** Fix the container case so an opaque type over a container carries its invariant into the prover, and let builtin signatures name a type defined in an Aver standard library. Then `Crypto.sha256 : Bytes -> Digest32` is expressible, with `Digest32` an ordinary refinement over `Bytes` instead of another bespoke type. Costs a standard library written in Aver, which is a new category here: loading, versioning, the self-host path. A compact runtime representation is separate work that has to land before any of this is usable on real payloads.

## Where I land

D, and I checked it was reachable before proposing it. Writing the Lean by hand in the shape the compiler would have to emit: the invariant survives, laws close with the tactics already in use, no `sorry`, only whitelisted axioms, nested refinements work, and an element pulled out of the container keeps its range. So it is buildable. It is also weeks of work rather than an afternoon.

I lean that way because of your use case. A transaction decoder wants several of these invariants, and C gives you none of them.

## What this means for your PRs

Under A nothing changes. Under B, C or D all four signatures change, including the merged `sendBytes`.

**None of this has shipped.** The last release is 0.27.1 from 19 July. `sendBytes` landed after it and isn't in the CHANGELOG yet, and the other three aren't merged. There is no published API to break and the only working tree affected is yours. That won't be true again once anything ships, so I'd rather settle it now than merge three more signatures on top of the problem.

## What I'm asking

You're the only person writing this code, so I'd rather hear your answer before I pick.

- Which option hurts your parser least, and where would each one bite?
- What does the decoder need beyond a byte sequence? Indexed access, fixed-length types for digests and magic values, a way to write a constant without spelling out four decimal numbers?
- If the signatures are changing, would you rather they change now, before your PRs merge, or land as they are with a migration afterwards?

jasisz:
Refs #786.

This lands the standard-library foundation for typed binary data:

- ships `Bytes` and `Crypto.Digest32` as ordinary Aver source embedded in the compiler
- keeps arbitrary byte sequences under `Bytes` and gives fixed-length digests their own `Crypto.Digest32` module
- adds total lowercase hex encoding and fallible case-insensitive hex decoding
- removes the old scalar `Byte.toHex` / `Byte.fromHex` namespace and its VM, Rust, WASM, Lean, Dafny, checker, LSP, and self-host wiring
- migrates the JSON codec and formal string-escape recognizer to `Bytes`
- makes standard modules installation-independent and reserved against project shadowing
- wires resolution through check, compile/proof, context, virtual filesystems/playground, and aver-lsp
- preserves both refinements through Lean and Dafny export
- re-canonicalises dependency expression TypeIds against the final whole-program symbol table, fixing nested carrier projection across module boundaries

Public shape:

- `depends [Bytes, Crypto.Digest32]`
- `Bytes.fromList : List<Int> -> Result<Bytes, String>`
- `Bytes.toList : Bytes -> List<Int>`
- `Bytes.fromHex : String -> Result<Bytes, String>`
- `Bytes.toHex : Bytes -> String`
- `Crypto.Digest32.fromBytes : Bytes -> Result<Digest32, String>`
- `Crypto.Digest32.toBytes : Digest32 -> Bytes`
- `Crypto.Digest32.fromHex : String -> Result<Digest32, String>`
- `Crypto.Digest32.toHex : Digest32 -> String`

`Digest32` is imported as the nominal type name; construction and projection live under the explicit `Crypto.Digest32` module path. Encoding is total because the `Bytes` invariant guarantees octets; decoding reports malformed input through `Result`.

Validation:

- VM JSON verification: 452/502 sampled cases, 0 failures, 50 skipped by `when`
- VM typechecker/eval suites: 471 passed
- wasm-gc standard-library fixture: 8/8
- Lean escaped-string roundtrip remains kernel-clean with zero `sorry`
- embedded `Bytes` / `Digest32` verify in both Lean and Dafny
- exact default and `wasm,wasip2` clippy commands with `-D warnings`

This intentionally does not close #786. Migrating byte-carrying service signatures such as TCP and adapting the existing SHA-256 PR remain follow-up work.


jasisz:
API naming follow-up: moved the digest refinement into its own embedded Crypto.Digest32 module. The public operations are now Crypto.Digest32.fromBytes and Crypto.Digest32.toBytes; consumers import depends [Bytes, Crypto.Digest32] and use Digest32 as the nominal type. Lean and Dafny end-to-end verification remain green.

PR #791, opened 2026-08-07
Make the wasm-gc effect test harness respect typecheck errors and fix the honest readBytes profile

jasisz:
## Problem

Three related gaps around `Tcp.readBytes` testing and modeling:

1. **The wasm-gc effect test harness ignored typecheck errors.** `run_wasm_gc_with_mode` in `tests/wasm_gc_effect_arg_overflow_regression.rs` runs the pipeline with `TypecheckMode::Full` but only read `result.analysis`, never the typecheck verdict. A program the compiler rejects could still be compiled and run by the harness.

2. **A test exercised a program Aver source cannot express.** `tcp_read_bytes_big_count_is_catchable_on_wasm_gc` constructed `Tcp.Connection(id = "missing", host = "", port = 0)` directly. `Tcp.Connection` is opaque, and feeding that exact source to the fixed harness fails with:

   ```
   [6] Cannot construct opaque type 'Tcp.Connection' — use its module's constructor function
   ```

   The test only passed because of gap 1.

3. **The "honest" hostile profile for `Tcp.readBytes` under-modeled the honest world.** The `normal_ok` stub returned `Bytes.fromList([0, 1, 2])` unconditionally, ignoring `count`. The real builtin is read-exact on every backend (`Ok` carries exactly `count` bytes, or `Err`), so `Ok` with 3 bytes for `count != 3` is impossible behavior — user code that validates frame length would never see a success path under the honest profile.

## Fix

- The harness now panics with the rendered errors when typechecking fails, matching `aver run --wasm-gc` (`try_run_wasm_gc`). All other tests in the file were audited: none relied on the gap.
- `tcp_read_bytes_big_count_is_catchable_on_wasm_gc` now obtains its connection from a real loopback listener via `Tcp.connect`, like the neighboring hosted TCP tests and the wasip2 sibling test, and still asserts that a 2^80 count surfaces as a catchable `Result.Err` (stdout `range-error`), not a trap. The listener holds the peer open until the guest closes so the error can only come from the count check.
- The `normal_ok` stub now honors the read-exact contract: `Ok` with exactly `count` zero bytes for `0 <= count <= 10485760` (the runtime's body limit), `Err` for negative counts, and `Err` for over-limit counts including out-of-i64 bignums, mirroring the builtin's error messages.

## Tests

- `cargo test --features wasm --test wasm_gc_effect_arg_overflow_regression` — 10 passed.
- `cargo test --features wasm hostile_effects` — 7 passed (includes the stub parse guard, the typecheck-vs-oracle-signature guard, and the profile-completeness guard).
- Verified the old forged source now trips the harness panic with the opaque-construction error (temporary probe, not committed).
- Smoke-ran a transcription of the new stub body through the VM: `count = 4` → `Ok` len 4, `count = 0` → `Ok` len 0, `count = -1` → `Err ... is negative`, `count = 2^80` → `Err ... exceeds the read limit`.
- `cargo fmt`; `cargo clippy --features wasm,wasip2 --tests` — no findings in touched files.

### DIFF
--- a/src/diagnostics/vm_verify.rs
+++ b/src/diagnostics/vm_verify.rs
@@ -476,7 +476,15 @@ pub(crate) fn inject_hostile_effect_stubs_for_blocks(
         })
         .collect();
 
-    let mut to_inject: Vec<&'static str> = Vec::new();
+    // Owned method names, deduplicated in declaration order. Owning the
+    // strings (instead of borrowing from `fn_def.effects`) is what lets
+    // the injection loop below take the mutable borrow of `items` — and
+    // it keys `hostile_profiles_for` on the classified method name
+    // verbatim, so a newly classified effect can never be silently
+    // dropped by a stale name mapping (the previous hand-maintained
+    // `&'static str` table mapped unknown methods to `""`, which made
+    // the Tcp byte methods' profiles vanish from this path).
+    let mut to_inject: Vec<String> = Vec::new();
     let mut seen: std::collections::HashSet<String> = std::collections::HashSet::new();
 
     for block in blocks {
@@ -498,12 +506,12 @@ pub(crate) fn inject_hostile_effect_stubs_for_blocks(
             if !seen.insert(method.to_string()) {
                 continue;
             }
-            to_inject.push(static_method_name(method));
+            to_inject.push(method.to_string());
         }
     }
 
     let mut new_items: Vec<TopLevel> = Vec::new();
-    for method in to_inject {
+    for method in &to_inject {
         for profile in hostile_profiles_for(method) {
             if existing.contains(&profile.stub_fn_name) {
                 continue;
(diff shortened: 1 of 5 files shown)


jasisz:
Follow-up commit addressing the review findings:

1. **Dead injection path (blocking)** — confirmed and fixed structurally. `inject_hostile_effect_stubs_for_blocks` now pushes owned method-name strings and keys `hostile_profiles_for` on the classified method name verbatim; the hand-maintained `static_method_name` table (whose `_ => ""` catch-all swallowed `Tcp.sendBytes`/`Tcp.readBytes`/`Tcp.writeBytes`) is deleted, so there is no mapping left to fall out of sync. Two guards:
   - a unit test iterating the whole `classify()` table: every non-Output classified method must get all of its profile stubs injected (red before the fix on all three byte methods);
   - an end-to-end test (`tests/regression_hostile_effect_stub_injection.rs`): a hostile law run over a fn with the `Tcp.readBytes` effect must fail exactly under the injected honest profile, whose successful read is impossible while the stubs are absent (an unstubbed dispatch can only fail on the fabricated connection). Red before the fix with 0 hostile failures.

   Audited the other `hostile_profiles_for` callers: the combination collector, `guard_for_case`, and `build_case_oracle_stubs` all use the method name verbatim, and the wasm-gc verify runner shares this same injector — no second mapping with the gap.

2. **Stub message fidelity (important)** — fidelity target is the VM builtin. The honest `normal_ok` stub now mirrors its branch order and text exactly: counts outside i64 (either sign) report `exceeds the read limit` before the sign check, in-i64 negative counts report `is negative`, and in-i64 over-limit counts name the limit (`exceeds the 10485760 byte limit`). `tests/hostile_readbytes_stub_fidelity.rs` compares the actual stub body (compiled through the VM) against the real builtin message-for-message on both divergent branches, plus pins the read-exact `Ok` length.

3. **Harness panic pin (minor)** — added a `should_panic(expected = "typecheck failed")` test feeding the harness the old ill-typed probe (opaque `Tcp.Connection` construction), so reverting the harness typecheck fix leaves a red test.

Note for a separate issue: while building the e2e test I hit a pre-existing typechecker rejection of `given reader: Tcp.readBytes = [stub]` — the oracle-signature comparison treats the table's `Bytes` and the stdlib's nominal `Bytes` as different named types, so the error prints two identical signatures. Not touched here; the e2e test uses a given-free law instead.

Ran: `cargo test --lib vm_verify` / `--lib hostile`, the two new integration suites, `cargo test --features wasm --test wasm_gc_effect_arg_overflow_regression` (11 passed), `cargo fmt`, `cargo clippy --features wasm,wasip2 --tests` (no findings in touched files).

PR #792, opened 2026-08-07
Resolve packed sequence layouts by exact type name

jasisz:
## Problem

The wasm-gc packed sequence lookups (`TypeRegistry::packed_sequence` in `src/codegen/wasm_gc/types.rs`, `packed_sequence_ops_lookup` in `src/codegen/wasm_gc/body.rs`, `PackedSequenceHelperRegistry::ops_for` in `src/codegen/wasm_gc/packed_sequences.rs`, introduced in #790) fell back to the bare last name segment when the exact key missed. The packed layout table (`packed_sequence_layout_table`) guards bare-name collisions via `ambiguous_names`, and the ungated-construction demotion scan matches names exactly — but the qualified-to-bare fallback sidestepped both guards.

Concrete failing example: the entry module declares a gated packed refinement candidate `Octets` (single `values: List<Int>` field, smart constructor gated by an all-in-`[0, 255]` predicate); dep modules `Left` and `Right` each declare a plain ungated `record Octets { values: List<Int> }`. The bare-name collision makes the multi-module flatten rename the dep records to `Left.Octets` / `Right.Octets`. The fallback then resolved `Left.Octets` to the entry refinement's packed u8 layout, so

```
plain = Left.Octets(values = [1000])
```

went through the pack helper and the packed run returned `232` (1000 mod 256) while `AVER_NO_PACKED_SEQUENCES=1` returned `1000` — silent truncation, no trap.

The scalar carrier-i64 path had the same hole: `is_eligible_carrier` / `is_eligible_carrier_field` (`src/codegen/wasm_gc/types.rs`) and the two projection-interval lookups in `src/ir/mir/optimize/bare_i64.rs` stripped qualified names to the bare segment. A collision-renamed plain record (`Left.IntRange`) inherited an unrelated gated carrier's i64 erasure and `Left.IntRange(value = 10000000000000000000)` trapped in `__aint_to_i64_checked` on the wasm path instead of round-tripping through the bignum representation (verified with the same two-dep repro shape).

## Fix

Exact-name matching only: remove the qualified-to-bare fallback from all of the lookup sites above. The layout / eligibility tables are keyed by post-flatten `TypeDef` names (bare for entry and non-colliding dep types, canonical `Prefix.Name` for collision-renamed dep types), and the collision guard plus the demotion scans match those names exactly, so lookups must too. A miss now keeps the ordinary boxed representation — the fail-closed direction. No alias registration turned out to be needed: every legitimate resolution site already sees the exact post-flatten name (the full wasm-gc, packed-sequence, carrier-i64 differential, stdlib Bytes, and Rust codegen differential suites pass unchanged).

## Tests

- New regression tests in `tests/wasm_gc_spec.rs` (multi-module harness): `cross_module_same_bare_name_record_does_not_inherit_packed_layout` (asserts `1000`, was `232`) and `cross_module_same_bare_name_record_does_not_inherit_carrier_i64_erasure` (asserts the beyond-i64 value survives, was a trap). Both reproduced the bug before the fix.
- `cargo test --features wasm --test wasm_gc_spec` — 39 passed
- `cargo test --features wasm --test wasm_gc_packed_sequence --test wasm_gc_codegen_regression --test stdlib_spec` — 12 passed
- `cargo test --features wasm --test wasm_gc_carrier_i64_differential` — 82 passed
- `cargo test --features wasm --lib bare_i64` — 43 passed; `--lib packed_sequence` — 4 passed
- `cargo test --test rust_codegen_differential` — 15 passed, 9 ignored
- `cargo fmt`, `cargo clippy --features wasm,wasip2` — no findings in touched files

### DIFF
--- a/src/codegen/wasm_gc/body.rs
+++ b/src/codegen/wasm_gc/body.rs
@@ -149,18 +149,14 @@ impl FnMap {
         }
     }
 
+    /// Exact-name lookup only — no qualified→bare fallback, mirroring
+    /// `TypeRegistry::packed_sequence`. A bare-name fallback would route
+    /// a collision-renamed dep type through an unrelated packed helper.
     pub(super) fn packed_sequence_ops_lookup(
         &self,
         type_name: &str,
     ) -> Option<super::packed_sequences::PackedSequenceOps> {
-        self.packed_sequence_ops
-            .get(type_name)
-            .copied()
-            .or_else(|| {
-                type_name
-                    .rsplit_once('.')
-                    .and_then(|(_, bare)| self.packed_sequence_ops.get(bare).copied())
-            })
+        self.packed_sequence_ops.get(type_name).copied()
     }
 
     pub(super) fn map_helpers_lookup(&self, canonical: &str) -> Option<&super::maps::MapKVHelpers> {
--- a/src/codegen/wasm_gc/packed_sequences.rs
+++ b/src/codegen/wasm_gc/packed_sequences.rs
@@ -53,12 +53,11 @@ impl PackedSequenceHelperRegistry {
         }
     }
 
+    /// Exact-name lookup only — no qualified→bare fallback, mirroring
+    /// `TypeRegistry::packed_sequence`. A bare-name fallback would route
+    /// a collision-renamed dep type through an unrelated pack/unpack pair.
     pub(super) fn ops_for(&self, type_name: &str) -> Option<PackedSequenceOps> {
-        self.ops.get(type_name).copied().or_else(|| {
-            type_name
-                .rsplit_once('.')
-                .and_then(|(_, bare)| self.ops.get(bare).copied())
-        })
+        self.ops.get(type_name).copied()
     }
 
     pub(super) fn iter(&self) -> impl Iterator<Item = (&str, PackedSequenceOps)> + '_ {
--- a/src/codegen/wasm_gc/types.rs
+++ b/src/codegen/wasm_gc/types.rs
@@ -1408,7 +1408,8 @@ impl TypeRegistry {
     /// ETAP-2 carrier-`i64`: install the set of refinement-via-opaque
     /// carrier type names whose proven bound `fits_i64`. Called once by the
     /// wasm-gc codegen entry after it derives the carrier interval table.
-    /// Keyed by bare Aver name (`"IntRange"`).
+    /// Keyed by post-flatten `TypeDef` name (`"IntRange"`, or the canonical
+    /// `"Left.IntRange"` for a collision-renamed dep type).
     pub(super) fn set_eligible_carriers(&mut self, names: std::collections::HashSet<String>) {
         self.eligible_carriers = names;
     }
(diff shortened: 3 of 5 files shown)


jasisz:
Follow-up audit of the open review question: "exact matching is safe only if post-flatten names are canonical everywhere." I enumerated every remaining bare-segment / qualified-to-bare matching site in `src/codegen` + `src/ir`, classified each as representation-affecting vs. pure name resolution, and ran concrete collision repros for the representation-affecting ones (entry-vs-single-dep twins, dep-dep renames, qualified spellings, container-element flow). Conclusion: **no remaining silent layout-inheritance hole; no code change needed.** Every non-canonical spelling now declines fail-closed (boxed representation or a compile-time error), never silently inherits a layout.

### Site-by-site verdicts

| Site | Class | Verdict | Why |
|---|---|---|---|
| `bare_i64.rs` `carrier_interval` / `field_carrier_field_interval` / `base_renders_eligible_carrier` | representation | safe | Exact match post-fix; table keys and MIR stamps are both post-flatten. |
| `bare_i64_rewrite.rs` `rewrite_for_rust` | representation | inert | Deliberately passes an EMPTY carrier table; carrier-i64 is wasm-gc-only. |
| `proof_lower/mod.rs` `carrier_interval_table` bare-key intersect | representation | benign today, flag | On the wasm path (flattened input, one scope) both same-name walks resolve identical info, so intersect degenerates to a no-op. The Rust backend computes this table with per-module scopes where intersecting on a name twin would *widen* `fits_i64` eligibility (the comment's rationale is inverted: intersection narrows the interval, which is not an over-approximation) — but `rewrite_for_rust` ignores the table, so it is unreachable. Revisit before ever enabling carrier-i64 on Rust. |
| `packed_sequence.rs` `same_callee_name` (suffix) in helper find + self-recursion check | representation | benign today, flag | Only production caller passes the flattened program; post-flatten fn names are underscore-joined (no dots), so the suffix branch cannot diverge from exact matching, and a dotted callee can only be a builtin, which cannot satisfy the smart-ctor shape. Latent hazard if the table is ever computed pre-flatten. |
| `common.rs` `refinement_info_for_in_scope` (first-match walk, `starts_with` on `Result<` payload) | representation | benign | Prefix match is guarded by the Ok-branch construct-name equality; the first-match across a post-flatten name twin resolves the entry ctor, and the ungated-construction demotion scan closes the divergence (verified, see repros). |
| `types.rs` `record_type_idx` / `sum_root_type_idx` / `record_field_index` / container lookups via `strip_inner_dotted_prefixes` (also `body.rs`) | slot lookup | benign, note | Bare fallback resolves the unique bare declarer; ambiguous shapes fail closed. Note a load-bearing inconsistency: `emit_mir_record_create` reads `registry.record_fields` with NO fallback while `record_type_idx` HAS one — that mismatch is what makes qualified constructs fail closed today (see repros). If someone later "harmonizes" it by adding the fallback to the field-list read, the truncation returns. Worth a comment or a regression pin in a follow-up. |
| `collect_lists_from_fn_body` / `collect_lists_from_expr` | name resolution | benign | Instantiation discovery; a miss surfaces as a codegen error, not a wrong layout. |
| `proof_lower/mod.rs` `find_fn_def_by_call_name` (+ Lean `shared.rs`, Dafny `lemmas.rs` twins) | name resolution | benign | Consumed only by Lean/Dafny proof emission, never by a representation decision. |
| `ir/hir/resolve.rs`, `ir/calls.rs`, `rust/from_mir.rs` `emit_mir_static_ref`, `codegen/common.rs` Rust-name mapping, `recursion/detect.rs` | name resolution | benign | Constructor-name splitting / Rust path mangling / recursion detection; failure mode is a compile error or a conservative decline. |

### Collision repros (wasm-gc multi-module harness)

- **Entry-vs-single-dep bare twin, dep plain, entry-side `Left.Octets(values = [1000])`** (flatten does NOT rename this shape — `colliding_bare_names` counts only dep-vs-dep declarers): pre-fix this silently returned **232**; post-fix it is a compile error (`record Left.Octets missing field list`). The fix closes this unrenamed-twin variant too, fail-closed. Scalar sibling: compile error both before and after (unchanged).
- **Entry-vs-single-dep twins, BOTH gated, dep bound wider** ([0,255] vs [0,65535] packed; [0,100] vs beyond-i64 scalar): correct results. Precision on the PR text: the `ambiguous_names` guard does NOT fire for this shape (both walks resolve the same first ctor, so the layouts compare equal) — the effective guard is the ungated-construction demotion scan, which sees the second gate's own `RecordCreate` and demotes the twin name. Verified both packed and scalar.
- **Dep-dep collision-renamed carrier inside `List` elements** (`[Left.IntRange(value = 10^19)]` alongside an eligible bare `IntRange` used in `List<IntRange>`): pre-fix this TRAPPED; post-fix correct. So the fix also covers container-element flow — a shape the PR's own regression tests don't pin, in case you want one more test.
- **Caveat (b), entry-side qualified local-binding annotation over a sole-declarer packed dep type** (`o2: Left.Octets = o` with `o` from the dep's gate): pre-fix this compiled and ran correctly (via the fallback); post-fix it is a wasm validation error — the local slot is typed from the pre-flatten STAMP, which keeps the qualified name, so the packed lookup declines to boxed while the value is packed. Fail-closed, but an availability regression on a legitimate program. The scalar sibling already failed validation pre-fix. Root cause: the flatten doc's claim that "stamped expression types use bare names" does not hold for entry-side annotations / cross-module sig payloads, so post-flatten names are NOT canonical in stamps. Restoring these shapes needs canonical stamping or a guarded alias table (qualified → bare only when the qualified name is not itself a typedef and the bare name has exactly one declarer) — not a blind fallback. Left as follow-up; nothing silent either way.

PR #793, opened 2026-08-07
Fix late build failure and carrier truncation in generated Rust for Crypto.sha256

jasisz:
## Problem

Two defects in the `Crypto.sha256` Rust codegen path (#781), plus a small dedup.

**1. Late build failure when `depends` omits `Crypto.Digest32`.** This program passes `aver check`, `aver verify`, and `aver compile --target rust`:

```
module Main
    intent = "Hash a byte sequence while depending only on Bytes."
    depends [Bytes]
    effects [Console.print]

fn hashTag(xs: List<Int>) -> Result<String, String>
    ? "Hash the octets and report that a digest was produced."
    bytes = Bytes.fromList(xs)?
    digest = Crypto.sha256(bytes)
    Result.Ok("hashed")

fn main() -> Unit
    ? "Hash a small byte list and print the outcome."
    ! [Console.print]
    match hashTag([1, 2, 3])
        Result.Ok(tag) -> Console.print(tag)
        Result.Err(msg) -> Console.print(msg)
```

and then the generated project fails `cargo build` with `error[E0433]: cannot find crypto in aver_generated` — the sha256 emission references `crate::aver_generated::crypto::digest32::Digest32`, but the digest32 module was only emitted when the user listed `Crypto.Digest32` in `depends`. The VM runs the same program fine; wasm-gc and wasip2 rejected it with ``builtin requires record `Digest32` `` even though the program is valid.

**2. Silent digest corruption on a malformed carrier.** The sha256 argument projection emitted `__b.to_u32().expect("Bytes invariant violated") as u8`. `to_u32()` returns `Some` for values in `256..2^32-1`, so `as u8` silently truncates and produces a wrong digest with no error. Only reachable on an already-violated `Bytes` invariant, but the neighboring Tcp byte builtins fail closed on the same condition.

## Fix

**1. Implicit inclusion of the stdlib modules a builtin's signature crosses** (the first of the two directions considered — chosen over a check-time diagnostic because it fixes all backends uniformly and keeps the valid program compiling instead of adding a new error for something the language accepts everywhere else). A single table in `src/stdlib.rs` maps each source-typed builtin to the standard modules that own its signature types (`Crypto.sha256` → `Bytes` + `Crypto.Digest32`; `Tcp.sendBytes`/`readBytes`/`writeBytes` → `Bytes`). The checker pass that canonicalizes those signatures (`canonicalize_source_typed_builtin_sigs`) now iterates the same table, so the builtin ↔ stdlib-type mapping has one home. At compile time, `load_compile_deps` / `load_module_recursive` scan each module for calls to these builtins and load the implied standard modules into the codegen module set — the same set every backend (Rust, wasm-gc, wasip2, VM lowering) already consumes. Explicit `depends` are unchanged and overlap is deduped; type checking and module visibility are untouched.

Verified with the real binary: the program above now builds and runs on Rust codegen (`cargo build` + run prints `hashed`), compiles and runs on wasm-gc and wasip2, and stays green on the VM. The explicit-`depends` fixture (`tests/fixtures/stdlib_bytes_app.av`) is unaffected.

**2. Fail-closed projection.** The generated projection is now `__b.to_u32().and_then(|__b| u8::try_from(__b).ok()).expect("Bytes invariant violated")` — a malformed carrier panics instead of truncating. sha256 is total (no `Result` channel), so panic is the only sound behavior, matching the VM which validates and errors.

**3. Dedup:** the VM's `Crypto.sha256` now calls the shared `aver_rt::crypto::sha256` wrapper instead of using `sha2` directly; the root crate's direct `sha2` dependency is dropped (aver-rt keeps it).

Known remaining gap (pre-existing, out of scope): the in-browser playground builds its module set through `load_module_tree_from_map` and does not apply the implicit inclusion, so the wasm-gc-in-browser path still requires explicit `depends`.

## Tests

- New regression `rust_sha256_builds_without_digest32_in_depends` in `tests/rust_codegen_differential.rs`: drives the failing program through `aver compile --target rust` + a real `cargo build` + run, asserting parity with the VM (fails with E0433 before the fix, passes after).
- New unit tests for the implicit-dependency scan in `src/stdlib.rs`.
- Ran locally: `cargo test --lib stdlib` (2 passed), `cargo test --lib checker` (203 passed), `cargo test --test stdlib_spec` (1 passed), `cargo test --test rust_codegen_differential rust_sha256_...` (1 passed) and `rust_tcp*` (3 passed), `cargo test --test rust_codegen_regression rust_codegen_builds_embedded_bytes_crypto_fixture` (1 passed), `cargo check -p aver-lang --lib --no-default-features --features playground`, `cargo fmt`, `cargo clippy --features wasm,wasip2` (clean).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -167,6 +167,7 @@ dependencies = [
  "proptest",
  "serde",
  "serde_json",
+ "sha2",
  "tempfile",
  "thiserror 1.0.69",
  "toml",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -167,7 +167,6 @@ dependencies = [
  "proptest",
  "serde",
  "serde_json",
- "sha2",
  "tempfile",
  "thiserror 1.0.69",
  "toml",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,6 +82,7 @@ wasm-bindgen = { version = "0.2", optional = true }
 wasm-encoder = { version = "0.248", optional = true }
 wasmparser = { version = "0.248", optional = true }
 wasmprinter = { version = "0.248", optional = true }
+sha2 = "0.10"
 wasmtime = { version = "44", optional = true, default-features = false, features = ["cranelift", "runtime", "gc", "gc-drc", "component-model", "component-model-async"] }
 wasmtime-wasi = { version = "44", optional = true }
 wasmtime-wasi-http = { version = "44", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,7 +82,6 @@ wasm-bindgen = { version = "0.2", optional = true }
 wasm-encoder = { version = "0.248", optional = true }
 wasmparser = { version = "0.248", optional = true }
 wasmprinter = { version = "0.248", optional = true }
-sha2 = "0.10"
 wasmtime = { version = "44", optional = true, default-features = false, features = ["cranelift", "runtime", "gc", "gc-drc", "component-model", "component-model-async"] }
 wasmtime-wasi = { version = "44", optional = true }
 wasmtime-wasi-http = { version = "44", optional = true }
--- a/src/call_graph.rs
+++ b/src/call_graph.rs
@@ -442,7 +442,7 @@ pub(crate) fn collect_callees_stmt(stmt: &Stmt, callees: &mut HashSet<String>) {
     }
 }
 
-fn collect_callees_expr(expr: &Spanned<Expr>, callees: &mut HashSet<String>) {
+pub(crate) fn collect_callees_expr(expr: &Spanned<Expr>, callees: &mut HashSet<String>) {
     walk_expr(expr, &mut |node| match node {
         Expr::FnCall(func, _) => {
             if let Some(callee) = expr_to_dotted_name(func.as_ref()) {
--- a/src/call_graph.rs
+++ b/src/call_graph.rs
@@ -434,7 +434,7 @@ fn count_recursive_calls_expr(expr: &Spanned<Expr>, recursive: &HashSet<String>,
     }
 }
 
-fn collect_callees_stmt(stmt: &Stmt, callees: &mut HashSet<String>) {
+pub(crate) fn collect_callees_stmt(stmt: &Stmt, callees: &mut HashSet<String>) {
     match stmt {
         Stmt::Binding(_, _, e) | Stmt::Expr(e) => {
             collect_callees_expr(e, callees);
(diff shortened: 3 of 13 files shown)


jasisz:
Follow-up commit addressing the review findings:

1. **Reverted the sha256 dedup in the interpreter.** `aver-rt` is an optional dependency behind the `runtime` feature, while `src/types/crypto.rs` compiles in feature combinations that do not enable it — verified with `cargo check -p aver-lang --lib --no-default-features --features certify`, which hit `E0433: cannot find module or crate aver_rt` at the dedup call site. `src/types/crypto.rs` calls `sha2` directly again (with a comment explaining why), the root `sha2` dependency is restored, and `aver_rt::crypto::sha256` remains for generated Rust projects only, which always link `aver-rt`.

2. **Implicit stdlib-dep loading moved into the shared loaders.** `source::load_compile_deps` / `load_module_recursive_for_compile` now include the modules implied by source-typed builtins, so `aver verify --wasm-gc` and the wasm-gc/wasip2 regression harnesses get the same behavior as the CLI loader. The scan itself lives in one place (`stdlib::implicit_stdlib_deps`); the command-local loader in `commands.rs` keeps its flag/error-handling shape but goes through that same helper. Red/green verified: with the shared-loader change reverted, the new wasm-gc regression fails with "builtin requires record `Digest32`"; with it, the suite passes.

3. **Verify blocks join the implicit-dep scan.** Rust codegen emits verify cases into a `#[cfg(test)]` module, so a program whose only `Crypto.sha256` call sits in a verify case used to build fine and then fail `cargo test` with E0433 on `aver_generated::crypto::digest32` (reproduced on the previous commit). The scan now covers verify case expressions and givens, and a new differential test compiles such a program and runs `cargo test` on the generated project. Note: the test avoids `?` inside verify cases — that trips a pre-existing, unrelated emitter gap (generated test fns return `Result<(), String>` while generated stdlib fns error with `AverStr`, so the `?` conversion fails E0277; reproducible on main with a plain `Bytes.fromList(...)?` verify case, no sha256 involved).

4. **Tcp byte-method and wasm-gc coverage.** New wasm-gc regression compiles a `depends []` `Tcp.readBytes` → `Tcp.writeBytes` relay (Bytes obtained only from the read) and asserts both host imports survive; the sha256-without-Digest32 wasm-gc test covers the shared-loader path for the crypto entry. Scan-level unit tests cover both table entries.

5. **Differential test comment + observation fixed.** The sha256 differential test now derives its printed output from the digests (structural equality of two hashes, agree/differ), so a future optimizer deleting an unused pure call cannot silently stop exercising the emission path; the comment states that risk.

6. **Playground routed through the same helper.** The refactor made this small: the virtual-fs loader (`load_recursive_from_map`) picks up implied modules for dep modules, and the playground's root-deps computation (`load_roots`) appends the entry's implied modules — all through `stdlib::implicit_stdlib_deps`.

Local runs: wasm-gc regression suite (10 passed), wasip2 regression (4), rust_codegen_differential fast tier (17), full `cargo test --lib` (957), CI feature-combo checks (`playground` lib check clean; `certify`-with-defaults unaffected), fmt + clippy clean on default, wasm,wasip2, and playground feature sets.

Issue #911, opened 2026-08-14
`Bytes.fromHex` decodes at about 0.8 MB/s compiled, because it allocates a `String` per hexadecimal character


n1bor:
`Bytes.fromHex` decodes at about 0.8 MB/s compiled, because it allocates a `String` per hexadecimal character

---

`Bytes.fromHex` is O(n) but with a very large constant: decoding a mebibyte
takes 1.24 seconds in a release build. It is written in Aver and explodes the
input into one boxed `String` per character before looking at any of them, so
every byte costs two string allocations, two `String.toLower` calls, and up to
thirty-two string comparisons.

This is not the quadratic-collection class of defect — the curve is straight.
It is the constant that hurts.

## Measured

Release build, `main` at `d6b72152`, timed with `Time.unixMs` around each step.
Sizes are the decoded output, so the hex input is twice each figure.

| output | `String.chars` alone | `Bytes.fromHex` | `Bytes.toList` |
|---|---|---|---|
| 256 KiB | 117 ms | 306 ms | 0 ms |
| 512 KiB | 227 ms | 625 ms | 0 ms |
| 1 MiB | 471 ms | 1,239 ms | 0 ms |

Linear, at roughly **0.81 MB/s of decoded output**. `String.chars` on its own is
38% of it. `Bytes.toList` is free, which is worth saying because it is the part
that looks expensive and is not — `Bytes` already wraps a list.

## Cause

`stdlib/bytes.av`:

```aver
fn fromHex(text: String) -> Result<Bytes, String>
    values = parseHexChars(String.chars(text), [])?
    fromList(values)
```

`String.chars` materialises a `List<String>` of 2n single-character strings for
an n-byte result. `parseHexChars` then walks it, and each character goes through

```aver
fn hexDigitValue(digit: String) -> Option<Int>
    match String.toLower(digit)
        "0" -> Option.Some(0)
        ...
        "f" -> Option.Some(15)
        _ -> Option.None
```

so per byte: two single-character `String` allocations, two `toLower` calls that
each allocate again, and two linear scans of a sixteen-arm string match. Then
`fromList` walks the result once more to range-check octets that the parse just
constructed from nibbles and cannot have put out of range.

`toHex` is the same shape in reverse — `byteToHex` builds a two-character
`String` per byte via `+`, and `hexParts` collects them for `String.join`.

## Where it hurts

I store Bitcoin blocks as hex text in
[...]

PR #930, opened 2026-08-15
Collect a list by appending to it instead of building it backwards

jasisz:
Part of #911.

A loop that collects with `List.prepend` and reverses once on the way out spends a cons cell per element and then walks all of them again, for a result whose elements arrived in the order it wanted. This gives such a loop a `<loop>__collected` variant that appends into a builder, and moves the call sites that start the accumulator at `[]` onto it. The original function stays exactly as written.

## The builder has no type of its own

It IS the `AverList<T>` the accumulator always was. The synthesized variant therefore has the same signature as the function it was built from, and neither a call site nor a backend has a new type to learn — `__lst_new` / `__lst_push` / `__lst_finalize` are `List<T> -> List<T>` throughout.

Compiled Rust writes into the backing vector in place. What makes that sound is the recogniser, not a runtime check: the accumulator is read exactly once on every path, so the list handed to a push is the only handle to what it holds. When it is not — a shape this is reused for later — the push copies and answers the same.

The virtual machine keeps immediate elements in a pool of its own and hands the builder back as the cons chain the moment a heap value arrives. A pool the collector cannot see must not hold anything the collector could move, and a growing arena object cannot be a tail-recursive loop's accumulator for free — the frame boundary rewrites everything a root reaches, so an N-element vector threaded through N iterations is rewritten N times. The cons chain is what the loop already paid for and what the boundary already handles in constant time per step, so that is the fallback: same allocations as before this pass, same answer, and the pool half keeps the win where the elements are numbers.

## Composition with the two rewrites before it

`Bytes.fromHex` reads a list of characters and writes a list of octets. The character list went in #928; the octet list goes here. What runs is `Bytes.parseHexChars__cursor__collected` — a cursor on one side, a builder on the other — which is the shape the fusion-ceiling probe measured by hand. The pass runs after chars fusion for exactly that reason, and `--emit-ir-after=list_build` is where you can see the second rewrite looking at the first one's output.

## Measurements

The #911 repro decodes hexadecimal text of the stated size with the untouched standard library. Serial, interleaved, `--release`, minimum of nine runs, with a do-nothing program's runtime subtracted so what is reported is the decode. Same machine, same session, one table so the issue can be answered from it.

**Compiled Rust — `Bytes.fromHex`, ms**

| input | pre-campaign (`59d2ec5`) | W1 (#918) | W3 (#928) | W5 (here) |
| 256 KiB | 50.7 | — | 26.0 | 9.1 |
| 512 KiB | 102.7 | — | 49.1 | 18.9 |
| 1 MiB | 203.0 | 182.2 | 95.3 | 37.0 |

**`aver run` — same repro, 1 MiB: 584 ms → 319 ms (W3) → 282 ms.**

**The list-build shape in isolation** — one million elements collected and reversed, nothing else in the way: compiled 120 ms → 13 ms, VM 133 ms → 46 ms.

The CHANGELOG quotes a separate end-to-end session measured during the review repair — whole program, wall clock, minimum of nine runs, nothing subtracted: 39 ms compiled and 300 ms on the VM for the 1 MiB decode, and 15 ms compiled for the isolated million-element loop. That session is not the runs in this table and the note no longer claims it is; the table stays the decode on its own, with the do-nothing subtraction.

So this wave is 2.6x on the repro and the campaign stands at 5.5x compiled. It does not reach the single-digit ms/MiB the wave was aimed at, and the emitted code says why: with the list side gone, what is left per output byte is two `AverStr` allocations for the one-character strings the cursor still materialises so `hexDigitValue` can take a `String`, plus the `Option` each digit is wrapped in. That is a cursor-side shape (a head that answers a codepoint), not a builder-side one.

## Declines

Every rule has a program in `tests/list_build_declines.rs` that answered correctly before this pass existed and still does: a loop that reads what it has collected, one that stops on what it has collected, one whose exits disagree about the reverse, one that restarts itself, one whose pattern re-binds the accumulator's name, one that hands the accumulator to a helper instead of returning it, a call site that starts with elements already in it, a caller that wanted the elements backwards, a program that binds `__lst_push` itself, and — added in review — a loop that leaves the reversing to its caller but can exit with a value the accumulator never reached, which previously fused and lost the caller's reverse on that path.

Each was checked by breaking its guard and watching it fail. Eight of the ten flip, the review-added untouched-exit guard among them. The two that do not are worth stating:

- **the restart** — leaving it alone would have answered correctly too, because the restart passes a fresh list to the loop the variant was built from, which is still in the program and still takes a list; it would only have abandoned the builder it was holding. The rule is here so that "the self-calls the rewrite classified" and "the self-calls the function has" are the same set by construction.
- **the shadowed name** — the VM masks it (its fallback builder prepends and reverses, which is the thing being replaced), and compiled Rust cannot host the shape at all: a cons pattern that binds a name a tail-call loop assigns to does not compile there today (`E0384`). That is a limitation of that backend, tracked, not fixed here. The rule is insurance for the day either changes.

Finding that is why two decline witnesses live in `rust_codegen_differential` instead: the VM alone cannot see their class of mistake.

## Proof invisibility

Registered below the `#925` snapshot next to `buffer_build` and `chars_fusion`, so nothing it invents can reach Lean or Dafny. `proof_snapshot_seam` asserts it on a fixture that fuses, non-vacuously — the pristine run must not fuse and the fused run must synthesize a variant, and both must emit the same proof text with no `__collected` or `__lst_` in it.

## Known, and not made worse

The wasm-gc `Intrinsic` arm falls through to `Ok(None)` for an intrinsic it does not know, rather than refusing. The three names added here take that same silent path if the pass's exclusion flag is ever wrong. The arm is untouched; the exclusion is checked instead — a fixture that fuses compiles to wasm-gc with none of the rewrites in the artifact and still verifies there.

## The second commit

`src/ir/buffer_build.rs` came out of the first commit at three and a half thousand lines, past the size that asks for a look. The second commit splits it along the seam it already had: `mod.rs` keeps the joined-builder pass and the recognisers both passes share, `list_build.rs` takes the collecting-loop pass built on them, and `tests.rs` takes both suites the way every other module in this tree keeps them. Pure move — the second pass reads its neighbours through `use super::*` exactly as it read them one screen up.

## Tested

`stdlib_spec`, `eval_spec`, `vm_spec`, `cross_backend_stress`, `rust_codegen_differential`, `proof_snapshot_seam`, `explain_passes_spec`, `buffer_build_acc_escape`, `chars_fusion_declines`, `list_build_declines`, `proof_spec`, `cert_certify_spec`, the wasm-gc and wasip2 suites, and the library unit tests. `cargo fmt --all -- --check` and the three `clippy -D warnings` invocations CI runs, including `--features wasm,wasip2`.


jasisz:
Repaired after review, two commits.

The critical one first: a loop that hands its accumulator back bare is paid for at the call site, where the rewrite removes the caller's reverse. An exit that returns some other value never owed that reverse, so any path through it came back un-reversed. The recogniser now counts such exits in the same walk that classifies where the reverse lives, and turns the loop down with its own reported reason ("an exit does not come from the accumulator, and the caller's reverse pays for the rewrite"). Loops that reverse inside their own exits keep accepting these exits — there the compensation sits inside exactly the exits that need it. The reviewer's program is now a regression test in tests/list_build_declines.rs: it answered [7, 8] before the fix and answers [8, 7] with it, and disabling the new check flips it back, so the test is holding the door. I also checked the other place a caller's reverse is removed — the String.join fusion — and it cannot have this hole: its recogniser only accepts a two-arm loop whose base arm is exactly the accumulator or its reverse, so there is no room for an exit of any other shape.

The minor one: the release note compared timings taken with the do-nothing subtraction against timings taken without it and called them the same runs, and the numbers could not reconcile. The note now quotes one fresh end-to-end session I ran for it — whole program, wall clock, minimum of nine runs: the one-megabyte decode at 39 ms compiled and 300 ms on the VM, the bare million-element loop at 15 ms compiled — and labels the earlier figures as a different session with an approximate comparison. The measurements footnote in this PR's description says the same.

PR #933, opened 2026-08-15
Pass a character's code to a classifier instead of a one-character string

jasisz:
Part of #911.

The chars-fusion cursor walks a string without building the list of its characters, but it still materialises each character as a one-character string, because the classifier it hands the character to — the standard library's `Bytes.hexDigitValue` — takes a `String`. That is two string allocations per output byte in the hex decode: one per character of the pair. This PR moves the character's **code** across the call boundary instead, and the last named per-byte allocation of the #911 repro goes away. The `Option` each digit is wrapped in stays; that is the boundary of this wave, not an oversight.

## Mechanism

Two halves, both inside the existing chars-fusion pass — no new pipeline stage.

**The classifier grows a codepoint variant.** After the codepoint-match rewrite, `hexDigitValue`'s whole body is one `match __str_code1_lower(digit)` with integer arms and a trailing wildcard. A callee of exactly that shape — one `String` parameter, no effects, not recursive, the parameter read nowhere but the match subject — gets a synthesized `<fn>__code(code: Int)` twin whose subject folds the codepoint directly: bare `code` for `__str_code1`, `__str_fold_lower(code)` / `__str_fold_upper(code)` for the folded forms. The `String` version stays in the program for every other caller.

**The peel binds the code.** Where the cursor loop bound `match __str_cursor_head(s, i)`, and every read of that head is either the sole argument of a qualifying classifier or something the rewrite can re-materialise, it now binds `match __str_cursor_code(s, i)` — no string exists for the character at all. Classifier calls become `<fn>__code(<code>)`; every other read is replaced in place by `__str_cursor_head(s, i)` **at the offset the binding read**, which matters because `parseHexChars` prints the offending character from inside a later peel, after the cursor has stepped twice. The error messages come out byte-identical — the differential pins them on the happy path, an odd-length input, an invalid ASCII character, and invalid multibyte characters in both the high and the low position.

Three new intrinsics back this, each one implementation in `aver-rt` with two thin call sites (VM opcodes 0xA2–0xA4, Rust codegen): `__str_cursor_code(s, i)` is the codepoint of exactly the character the head read would produce, and `__str_fold_lower` / `__str_fold_upper` are the codepoint-level case folds. The folds are pinned by an exhaustive test over **all 1,114,112 Unicode scalar values** against the string-level folds they stand in for — `U+212A KELVIN SIGN` keeps its `"k"` arm, `U+0130` keeps the wildcard, surrogates and out-of-range codes answer `-1`.

## Guards and declines

- A classifier that reads its parameter anywhere outside the subject grows no variant — its `__code` twin would keep a read with no string behind it. The decline suite carries a program where that dropped read would have silently resolved to a top-level binding of the same name: breaking the guard flips the answer from `letterq` to `letterzzz`.
- A recursive classifier (tail calls included — those hide their target from an identifier walk), one with effects, more than one parameter, or a body that is not exactly the rewritten match: no variant. Opportunity detection, not a decline — nothing is reported.
- A call whose argument is not the head binder keeps the `String` classifier, which stays in the program precisely for that.
- The peel's code binder is fresh per peel. Nested peels each bind their own code, and a shared name would hand the outer classifier the **inner** character's code — injected exactly that fault: with the freshness broken and the shadow guard intact, the loop declines and stays correct (`declined Bytes.parseHexChars: a binder between the head binding and a read shadows a cursor name`); with the guard also disabled, `00ff10AbCdEf` decodes to `00ff00bbddff`. The guard is unreachable through the shapes the pass itself emits today and says so in its doc.
- `<fn>__code` name collisions decline the variant through the same program-wide taken-names set the cursor variant uses; a variant no rewrite reached is dropped rather than shipped as dead code.
- **A binder spelled like any name the pass emits declines the loop whole** — `__str_cursor_head`, `__str_cursor_code`, `__str_cursor_next`, `__str_cursor_end`, the `__str_code1*` family, `__str_fold_*` — through the same taken-name decline that already covered `__cur_`. The emitted calls are bare identifiers resolved after this pass, so a parameter, statement binding, pattern binder, or a single-arm match binding the name between a head's binding and its read would capture them: fused, a loop whose error arm bound `__str_cursor_head` and interpolated the head crashed on the VM (`cannot call non-function … in walk__cursor`) where the same program under any other binder name prints `bad x at 1 / bad é at 0 / done 2`; a loop **parameter** named `__str_cursor_head` died with `Runtime effect violation: cannot call 'Args.get'` instead of printing `done 3` — that one reproduces on main, the same hole a wave older. The codepoint-match rewrite emits `__str_code1*` into whatever function a single-character match sits in, loop or not, so it now skips a function that binds into the namespace. The regression tests walk the whole emitted namespace across all four binder positions in one loop, with an ordinary name in the same loop as the control.
- **The buffer-build pass had the same hole under its own names** (also reproduces on main): a caller that bound `__buf_finalize` captured the finalize call the join-site rewrite emits, and a collecting loop whose element binder was spelled `__sep` silently handed the buffered variant's separator read the loop's element — `a,b,c` came out `abbcc`. It now steps aside whole when the program binds any `__buf_*` name or the `__buf` / `__sep` binders its variants introduce — the same program-wide rule the list-build pass already applied to `__lst_`. List-build itself needed nothing: its existing guard reads the same all-binder-positions name set.

## The two backends and the proof line

The VM re-parses dependencies and adopts the fused copy only when the entry's symbol table knows every synthesized name — `Bytes.hexDigitValue__code` rides the same path `parseHexChars__cursor` already rides, and a new bytecode-level test pins the adoption: the compiled decoding loop contains `STR_CURSOR_CODE` and the classifier variant contains `STR_FOLD_LOWER`. The list-build pass composes on top unchanged — what runs is `parseHexChars__cursor__collected`, cursor-codes on one side, builder on the other. The new intrinsic names all start with `__str_`, so the existing proof-seam assertion (`proof_snapshot_seam`) covers them for free; wasm-gc and wasip2 are excluded by the same flags as the rest of the family and their artifacts carry none of this.

## Measurements

One fresh session, one method: whole-program wall clock, nine runs after one excluded warmup, min and median, nothing subtracted. Release builds. **Before** is the merge base `b8b895da` (which already carries the whole W1–W5 family); **after** is this branch. Two programs: the #911 repro (decode 1 MiB of hexadecimal with the untouched standard library, print the byte count) and the isolated shape (a cursor loop handing each of two million characters to a sixteen-arm classifier behind `String.toLower`).

| program | backend | before min / median | after min / median | change |
| 1 MiB hex decode | compiled Rust | 75.6 / 77.0 ms | 27.3 / 28.0 ms | ~2.8× |
| 1 MiB hex decode | `aver run` | 580.6 / 586.2 ms | 562.2 / 565.7 ms | ~3–4% |
| classifier loop, 2M chars | compiled Rust | 54.9 / 55.9 ms | 13.6 / 15.4 ms | ~3.6–4× |
| classifier loop, 2M chars | `aver run` | 300.4 / 301.8 ms | 283.9 / 286.1 ms | ~5–6% |

The change column spans the min-pair and the median-pair ratio wherever the two round apart. The compiled side is where the two `AverStr` allocations per output byte lived, and removing them is worth 2.8× on the repro. The VM barely moves: its per-step cost is opcode dispatch, and an arena string of at most four bytes was never its bottleneck — stated as measured, not promised. These numbers are whole programs (the input is built by string doubling inside the program), so they are not comparable to the W5 table, which subtracted a do-nothing baseline; the CHANGELOG quotes exactly this session.

## Red receipts

- `a_classifier_call_in_a_cursor_loop_moves_to_the_codepoint`, `a_cold_read_rematerialises_the_head_at_the_binds_own_offset` (the order pin — the re-materialised head must read the bind-time offset with a later peel's offset in scope), and `a_classifier_argument_that_is_not_the_head_keeps_the_string_call` all failed before the pass existed (`codepoint_calls: 0`, no `__code` in `synthesized`) and pass after.
- Fault injection, param-read guard: removed → `letterq` becomes `letterzzz` → restored.
- Fault injection, freshness/shadow guard: non-fresh binder with guard on → reported decline, correct output; guard off → `00ff10abcdef` becomes `00ff00bbddff` → both restored.
- Fault injection, binder-namespace guards: all three disabled → every namespace-walk test fails — the parameter witness with `Runtime effect violation: cannot call 'Args.get'`, the between-bind-and-read witness with `cannot call non-function … in walk__cursor`, the buffer sink witness with `cannot call non-function … in parts__buffered` — → restored, all seven green.

## Tested

`chars_fusion` unit tests, `chars_fusion_declines`, `list_build_declines`, `buffer_build_acc_escape`, `rust_codegen_differential`, `eval_spec`, `hir_mir_differential`, `proof_snapshot_seam`, `stdlib_spec`, `explain_passes_spec`, `vm_spec`, `mir_vm_codegen`, `cross_backend_stress`, `proof_spec`, and the `aver-rt` unit tests with the exhaustive fold receipt. `cargo fmt --all -- --check` clean; `cargo clippy --features wasm,wasip2` clean on the changed files.


jasisz:
Correction to the sibling-pass paragraph: the namespace guards added here cover the three fusion passes, but string interpolation (interp_lower) still emits __buf_*/__to_str calls unguarded, and the intrinsic-over-user-fn direction is open everywhere. Both are pre-existing on main and tracked in #938.

Issue #992, opened 2026-08-18
A re-exposed type from a dependency resolves for the root module and not for a module reached as a dependency


n1bor:
A module that lists a dependency's type in its own `exposes` re-exports it correctly — but only for the module being checked directly. The same consuming module, reached as a dependency of something else, is refused.

## Repro

Three modules. `Low` declares the type, `High` re-exposes it, `Mid` uses it in a signature.

`low/low.av`

```aver
module Low
    intent = "Owns a record that a module above wants to hand on under its own name."
    exposes [Item, sized]
    depends []
    effects []

record Item
    key: String

fn sized(item: Item) -> Int
    ? "How long the key is."
    String.len(item.key)

verify sized
    sized(Item(key = "ab")) => 2
```

`high/high.av`

```aver
module High
    intent = "Re-exposes a type it depends on, so its own callers need not know."
    exposes [Item, doubled]
    depends [Low.Low]
    effects []

fn doubled(item: Item) -> Int
    ? "Twice the key length."
    Low.Low.sized(item) * 2

verify doubled
    doubled(Item(key = "ab")) => 4
```

`mid/mid.av`

```aver
module Mid
    intent = "Consumes the re-exposed name while itself being a dependency, not the root."
    exposes [tripled]
    depends [High.High]
    effects []

fn tripled(item: Item) -> Int
    ? "Half again as much as the doubled length."
    High.High.doubled(item) + 1

verify tripled
    tripled(Item(key = "ab")) => 5
```

`main.av` depends on `Mid.Mid` and calls `tripled`.

Checked on its own, `Mid` is fine:

```
$ aver check mid/mid.av --module-root .
  ✓ types | 12 lines
```

Checked as a dependency of `main.av`, the identical file is refused:

```
$ aver check main.av --module-root .
error[type-error]: Function 'tripled', parameter 'item': Unknown type 'Item' — no type of that name is declared in this module or exposed by one it depends on
error[type-error]: Argument 1 of 'Mid.Mid.tripled': expected Item, got Low.Low.Item
```

The second line is the interesting one: the checker resolved the *argument* to `Low.Low.Item` and the *parameter* to nothing, so the two halves of the same call disagree about whether the re-export exists.

Adding `Low.Low` to `Mid`'s `depends` — a module `Mid` neither names nor needs — makes it pass. That is the workaround, and it defeats the point of the re-export.

## Why it matters

[...]

Issue #883, opened 2026-08-13
Map equality still diverges from runtime for key types with no ordering


jasisz:
Two maps built from the same entries in different orders are equal when the program runs and unequal in the exported Lean model, so a `verify` block can pass and its exported proof can be refuted by the kernel. The same applies to `!=` and to a map comparison used as a `when` premise.

Measured, with plain `String` keys:

```
fn aMap() -> Map<String, Int>
    ? "Insert x then y."
    Map.set(Map.set({}, "x", 1), "y", 2)

fn bMap() -> Map<String, Int>
    ? "Insert y then x."
    Map.set(Map.set({}, "y", 2), "x", 1)

fn same() -> Bool
    ? "Both orders build the same map."
    aMap() == bMap()

verify same
    same() => true
```

`aver verify` → `1/1 cases passed`. `aver proof` then `lake build` →

```
error: Main.lean:21:28: Tactic `decide` proved that the proposition
  same = true
is false
```

Mechanism: at run time a map is a hash map and equality is by content. In the model, `AverMap` is an association list (`src/codegen/lean/prelude.rs`, emitted into `AverCommon.lean`), `AverMap.set` appends an absent key at the end, and the emitted `==` is list equality — positional, so the two build orders are different values. Model equality is strictly finer than the language's.

Note this is wider than `Float` or non-scalar keys; the probe above uses `String`.

A fix has to establish that map equality in the model quotients over entry order — a normalised comparison or a set-like carrier — or that a map comparison is refused at export so `verify` and `proof` cannot reach opposite verdicts. What must not survive is a passing `verify` case exporting to a goal the kernel disproves.

jasisz:
Narrowing this: it was written against the tree before #874, which closes most of it.

#874 makes the Lean model canonically sorted — `AverMap.set` inserts in key order and `fromList` folds through `set` — so for `Int`, `String` and `Bool` keys list equality now coincides with the extensional equality every executable backend implements. Two maps built by inserting the same pairs in a different order are equal in the model, as they already were at runtime.

What survives, and it is the more interesting half:

- Key types with no `AverKeyOrder` instance fall to the priority-50 fallback `⟨fun _ _ => false⟩`, which makes `set` append again. `Float` and any non-scalar key are in that set. For those the model is still finer than runtime, so Lean can prove two extensionally-equal maps distinct.
[...]

PR #1005, opened 2026-08-19
fix: require an ordering on map key types

jasisz:
## Summary

- admit `Int`, `String` and `Bool` as map key types and refuse the rest at check time, wherever the key type is decided
- close the path where a passing `aver verify` and a certified `aver proof` disagreed about a map
- keep the export refusal as a backstop, with its own list, and hold that list against the language's and against the Lean model's orderings
- move the fixtures whose subject can no longer be written to expected-error tests, and keep the cone-walk fixtures over a map whose key stays invisible to the exporter

## The hole this closes

`aver proof` refuses to export a claim whose key type it cannot order, but it read that key type from every signature the claim reached rather than from the map actually being read. One modelled key anywhere in reach opened the gate for all of them. Measured on `main` (`15a2ee4c`):

```aver
fn keysOf(s: Map<String, Int>) -> List<Float>   # the parameter is never used
    m = {2.0 => 2, 1.0 => 1}
    Map.keys(m)

verify keysOf
    keysOf({}) => [2.0, 1.0]
```

```
aver verify → 0/1 cases passed | 1 failed   (actual: [1, 2])
aver proof  → exit 0, build_errors 0, sorries 0, nothing declined
              example : keysOf [] = [2.0, 1.0] := by native_decide
```

The kernel certified what `verify` refuted. Removing the unused parameter was enough to make the same file decline. That program no longer typechecks.

## Doors

The rule is one predicate (`src/types/map_key.rs`) called from every place a key type is decided: annotations (the `Type::Map` arm of `reject_fn_in_type`, which carries parameters, return types, record fields, variant payloads and binding annotations), map literals, both `Map.set` inference paths, `Map.fromList`, a read of an empty map (`Map.has({}, 1.0)` decides `K` at the call), `verify` givens, capability operation signatures, and signatures registered from a dependency.

That last group matters because a dependency's body is only re-checked when a module root is given; without it, registration is the only place the type is seen.

## What is deliberately not in scope

The export refusal in `src/codegen/common.rs` stays. Under the ban its named-key branches are unreachable from source, and that is the point of a backstop — it now keeps its own list rather than reading the language's, so widening one is a deliberate decision to widen the other. A new test holds the two lists and the Lean prelude's `AverKeyOrder` instances together: adding a fourth key type to the language without an instance in the model fails it.

`Refs #883` rather than `Fixes`: the remaining half of that issue is wasm-gc iteration order, which returns hash-bucket order even for `String` keys, so the ban does not touch it.

## Cost

Outside `tests/`, nothing in the repo keys a map on anything but `String` — 181 × `Map<String, Int>`, 84 × `Map<String, Val>`, and so on, with zero `Float`, tuple or list keys in `stdlib/`, `examples/`, `projects/`, `self_hosted/` or `challenges/`. The reporter's own project (`n1bor/btc-listener`, 61 `.av` files) has 17 maps, all `Map<String, String>`; it checks clean with `--deps`.

`Map<Float, …>` and `Map<List<…>, …>` passed `aver check` before this change and then failed to compile in generated Rust (`f64: Eq` / `AverList: Ord` not satisfied), so the ban turns an rustc error into an Aver diagnostic rather than removing something that worked.

## Also in this PR

A missing `CHANGELOG` entry for the dependency type re-export fix (#992/#1000), which shipped without one.

## Testing

- `cargo clippy --workspace --all-targets --exclude aver-lang -- -D warnings`
- `cargo clippy -p aver-lang --lib --bin aver -- -D warnings`
- `cargo clippy -p aver-lang --lib --bin aver --features wasm,wasip2 -- -D warnings`
- `cargo test --workspace` (3290 passed, 27 ignored)
- `cargo run --bin aver -- check examples/core`
- `cargo run --bin aver -- check examples/data`
- the reporter's project, `aver check main.av --module-root . --deps`


jasisz:
Superseded by #1006, which fixes the same hole without narrowing what a map can be keyed on.

This one refused every key type the proof model does not already order, which took records and variants with it — including the newtype pattern the language teaches, and `Map<Octets, …>` in our own wasm-gc suite. That was cutting the language to avoid writing a comparator.

#1006 writes the comparator instead: composite keys get one canonical order stated the same way on the VM and in compiled Rust, only `Float`, `Map` and `Vector` are refused, and the export gate reads the key type of the map a claim observes rather than of every signature its cone mentions — which is what let an unused parameter open it.

Issue #1065, opened 2026-08-21
The borrowed-equality fix misses the packed AverIntList representation


n1bor:
Follow-up to jasisz/aver#1037, which is fixed for an ordinary `AverList<AverInt>` — I confirmed that on current `main` and withdrew jasisz/aver#1060 accordingly. It is **not** fixed for the packed `AverIntList` representation that 8af74ec8 (*perf(rust): carry proof-packed bytes through `List<Int>`*) introduced.

## What fails

`n1bor/btc-listener`, `domain/witness.av`:

```aver
fn matchedScript(program: List<Int>, script: List<Int>, stack: List<List<Int>>, context: Context) -> Outcome
    ? "The witness Script only runs if it is the one the Output committed to."
    match sha256Of(script) == program
        false -> Outcome.Decided(Ending.Failed("…"))
        true -> executed(script, stack, context)
```

`aver check` clean, `aver verify` 75/75 on the file, and then:

```
error[E0308]: mismatched types
   --> src/aver_generated/domain/witness/mod.rs:199:69
    |
199 |     if (crate::…::witness::sha256Of(script) == program) { … }
    |         -----------------------------------    ^^^^^^^ expected `AverIntList`, found `&AverIntList`
    |         |
    |         expected because this is `AverIntList`
    |
help: consider dereferencing the borrow
    |
199 |     … sha256Of(script) == *program …
```

Same shape as #1037 — call result owned on the left, parameter borrowed on the right — but the type is `AverIntList` rather than `AverList<AverInt>`, so whatever the #1038 fix keys on does not reach it.

## What I could not reproduce

I tried three times to shrink this and **all three compile clean**, so the packed representation is not chosen in any of them:

* a `List<Int>` returned from a plain literal (the original #1037 repro)
* a `List<Int>` from `Bytes.toList(Bytes.fromHex(…))`
* a `List<Int>` from `Bytes.toList(Crypto.Digest32.toBytes(Crypto.sha256(…)))` — which is exactly what `sha256Of` does in the failing module

So the trigger is something about the surrounding module rather than the expression, and you will find it faster than I will — hence the generated line above rather than a repro file. The real one is `domain/witness.av` in n1bor/btc-listener at commit 002db47, `matchedScript` with the `sameBytes` helper inlined.

## Why I am filing it rather than just working around it

The workaround is trivial — bind both sides as parameters:

```aver
fn sameBytes(left: List<Int>, right: List<Int>) -> Bool
    left == right
```

[...]

Issue #1060, opened 2026-08-21
Comparing a list-typed call result with a parameter still emits Rust that does not compile, in both operand orders (#1037 reopened)


n1bor:
jasisz/aver#1037 is closed but still reproduces on **aver 0.28.1**, and in both operand orders — the original repro had the call on the left, and putting the parameter there fails too, with a different error. I could not reopen that issue, so this is a fresh one; a comment with the same evidence is on it.

## Repro

```aver
module Main
    intent =
        "Minimal repro: comparing a parameter with a call result, both orders."
    exposes [main]
    depends []
    effects [Console.print]

fn made() -> List<Int>
    ? "A freshly built list, returned by value."
    [1, 2]

fn callFirst(other: List<Int>) -> Bool
    ? "Call on the left, parameter on the right."
    made() == other

verify callFirst
    callFirst([1, 2]) => true
    callFirst([3]) => false

fn paramFirst(other: List<Int>) -> Bool
    ? "Parameter on the left, call on the right."
    other == made()

verify paramFirst
    paramFirst([1, 2]) => true
    paramFirst([3]) => false
```

```
$ aver verify main.av --module-root .
  ✓ callFirst      2/2
  ✓ paramFirst      2/2

$ aver compile main.av --module-root . -o out && (cd out && cargo build --release)
error[E0308]: mismatched types
  --> src/aver_generated/entry/mod.rs:13:16
   |      ------    ^^^^^ expected `AverList<AverInt>`, found `&AverList<AverInt>`

error[E0277]: can't compare `&aver_rt::AverList<aver_rt::AverInt>` with `aver_rt::AverList<aver_rt::AverInt>`
  --> src/aver_generated/entry/mod.rs:19:12

error: could not compile `main` (bin "main") due to 2 previous errors
```

Line 13 is `callFirst`, line 19 is `paramFirst`. Two different diagnostics for the same defect: one side of the comparison is emitted as a reference and the other by value, and which is which depends on the order. rustc's own suggestion — `*script` — is what the emitter should be doing.

## Where it bit

`n1bor/btc-listener`, classifying Core's pay-to-anchor Output. That Output is `OP_1` and a push of the two bytes `0x4e73`, four bytes in full, so the natural way to write the test is to compare against those four bytes:

```aver
fn anchorOrWitness(script: List<Int>) -> String
    ? "Core's pay-to-anchor, which is one particular version one program."
    match script == anchorScript()
        true -> anchorName()
        false -> witnessName()
```

[...]

Issue #1037, opened 2026-08-21
Comparing a call result with a parameter verifies but emits Rust that does not compile


n1bor:
`aver verify` passes and `cargo build` on the emitted Rust fails, for an equality between a call result and a parameter of the same list type.

## Repro

```aver
module Main
    intent =
        "Minimal repro: comparing a call result with a parameter."
    exposes [main]
    depends []
    effects [Console.print]

fn made() -> List<Int>
    ? "A freshly built list, returned by value."
    [1, 2]

fn same(other: List<Int>) -> Bool
    ? "Compare what a call returns with what a parameter holds."
    made() == other

verify same
    same([1, 2]) => true
    same([3]) => false

fn main() -> Unit
    ? "Say what it found."
    ! [Console.print]
    Console.print("{same([1, 2])}")

verify main
    main() => Unit
```

```
$ aver verify main.av
  ✓ same      2/2

$ aver compile main.av -o out && cd out && cargo build --release
error[E0308]: mismatched types
   |      ------    ^^^^^ expected `AverList<AverInt>`, found `&AverList<AverInt>`
```

The emitted line is

```rust
pub fn same(other: &aver_rt::AverList<aver_rt::AverInt>) -> bool {
    (crate::aver_generated::main::made() == other)
}
```

`made()` returns an owned `AverList<AverInt>`, `other` is `&AverList<AverInt>`, and `PartialEq` is not implemented across that pair. `rustc` itself suggests the fix: `== *other`.

## Why it is not seen more often

The same shape compiles when the function ends up in a tail-call group. `Domain.ScriptParse.keptUnlessMatched` in n1bor/btc-listener does exactly this:

```aver
fn keptUnlessMatched(op: Op, rest: List<Op>, target: List<Int>, acc: List<Op>) -> List<Op>
    match bytesOf([op]) == target
```

and is fine, because it is emitted as a trampoline entry that `.clone()`s every parameter to an owned value first, so the comparison never sees a borrow:

```rust
pub fn keptUnlessMatched(op: &Op, rest: &AverList<Op>, target: &AverList<AverInt>, acc: &AverList<Op>) -> AverList<Op> {
    __mutual_tco_trampoline_2(__MutualTco2::KeptUnlessMatched(op.clone(), rest.clone(), target.clone(), acc.clone()))
}
```

[...]

PR #1107, opened 2026-08-23
Tail-call lowering: keep loop-carried values alive, and align mutual-trampoline equality

### LINKED ISSUE #1105: Rust backend at 9d4131de: tail-call-lowered functions move a carried parameter, and == between a temporary and a parameter misses a deref
## What we are building

n1bor/btc-listener, moving its pin from ea1bfa56 to 9d4131de to take #1099 and #1104. `aver check` and `aver verify` pass (6,277 cases, in 19 s — thank you); `aver compile` then produces Rust that does not build: **7 errors in 6 modules**, every one in a tail-call-lowered function, none of which changed since the last pin. At ea1bfa56 the same sources compiled and ran a 400,000-Block audit.

## Two shapes, reduced to 40 lines

```aver
module Main
    intent =
        "Reproducer: two tail-call shapes the Rust backend at 9d4131de no longer"
        "lowers to code that compiles."
    exposes [main]
    depends []
    effects [Console.print]

fn tagged(tag: String, items: List<Int>, index: Int, acc: List<String>) -> List<String>
    ? "A parameter that is both carried into the self tail call and used in it."
    match items
        [] -> acc
        [head, ..tail] -> tagged(tag, tail, index + 1, List.prepend(label(tag, index, head), acc))

verify tagged
    tagged("t", [7, 8], 0, []) => ["t:1:8", "t:0:7"]

fn label(tag: String, index: Int, head: Int) -> String
    ? "The carried parameters, used."
    "{tag}:{index}:{head}"

verify label
    label("t", 0, 7) => "t:0:7"

fn without(items: List<List<Int>>, target: List<Int>, acc: List<List<Int>>) -> List<List<Int>>
    ? "Mutual tail recursion whose partner compares a computed list to a parameter."
    match items
        [] -> List.reverse(acc)
        [head, ..tail] -> kept(head, tail, target, acc)

verify without
    without([[1], [2]], [1], []) => [[2]]

fn kept(item: List<Int>, rest: List<List<Int>>, target: List<Int>, acc: List<List<Int>>) -> List<List<Int>>
    ? "An owned temporary on the left of ==, a parameter on the right."
    match List.reverse(item) == target
        true -> without(rest, target, acc)
        false -> without(rest, target, List.prepend(item, acc))

verify kept
    kept([1], [], [1], []) => []

fn main() -> Unit
    ? "Print both results."
    ! [Console.print]
    _a = Console.print(String.join(tagged("t", [7, 8], 0, []), ","))
    Console.print("{List.len(without([[1], [2]], [1], []))}")
```

`aver verify main.av` → 4/4. `aver compile main.av -o out && cd out && cargo build` →

```
error[E0308]: mismatched types
  --> src/aver_generated/entry/mod.rs:19:39
19 | ...   if (item.reverse() == target) { __MutualTco1::Without(rest, acc) } else { ...
   |           --------------    ^^^^^^ expected `AverIntList`, found `&AverIntList`

error[E0382]: use of moved value: `tag`
  --> src/aver_generated/entry/mod.rs:43:59
37 | pub fn tagged(mut tag: AverStr, mut items: aver_rt::AverIntList, mut index: aver_rt::AverInt, mut acc: ...
   |               ------- move occurs because `tag` has type `aver_rt::AverStr`, which does not implement the `Copy` trait
38 |     loop {
   |     ---- inside of this loop
```

1. **A carried parameter, used.** A self tail call that passes `tag` through unchanged *and* hands it to another function in the same arm. The loop lowering moves `tag` into the call and then needs it again on the next iteration; the old backend cloned, the new one does not. In btc-listener this is `infra/outputs.av entriesFor` (`txid`), `infra/txindex.av sitesFor` (`blockId`), `domain/inventory.av entriesFor` (`kind`), `domain/bech32.av checksumDigits` (`residue`), `domain/scriptwork.av spreadOver` (`position`, `weight`) — Strings and Ints alike.
2. **An owned temporary compared to a parameter.** In a mutual-recursion group, `List.reverse(item) == target` lowers to `item.reverse() == target` where `target` is `&AverIntList`; rustc wants `*target`. In btc-listener: `domain/scriptparse.av keptUnlessMatched`, which is FindAndDelete.

Bisecting by eye, cedafe0b (#1098, "finish resolved-IR migration and remove AST/codegen bridges") or 2e009dd7 (#1091, "consume resolved functions in codegen") are the candidates; I have not built the intermediate commits.

## What we are not asking for

Not asking for the loop lowering to change shape — the tail-call-to-loop is what keeps a 400,000-iteration walk off the stack, and the clone it used to insert is the right price. Only that it compile again.

jasisz:
Fixes #1105. Two shapes, two separate causes, both reported against 9d4131de by n1bor/btc-listener, whose sources had not changed since the previous pin.

Regression introduced by cedafe0b (#1098, the resolved-IR migration), found by bisect with an independent rebuild of the good and bad commits rather than by reading the diff.

## Shape 1 — a loop-carried parameter was moved

```aver
fn entriesFor(txid: String, items: List<Int>, index: Int, acc: List<String>) -> List<String>
    match items
        [] -> List.reverse(acc)
        [head, ..tail] -> entriesFor(txid, tail, index + 1, List.prepend(entryFor(txid, index), acc))
```

Resolved pass-through detection elided the identity TCO argument for `txid`, so the last-use analysis let the helper call move it — not seeing the use hidden in the next lap of the loop. `error[E0382]: use of moved value`.

Loop-carried parameters are now tracked per tail call, and the existing ownership policy clones only at genuine owning uses. Identity pass-through, borrows and `Copy` values stay clone-free — a blanket clone here would be a silent performance regression on a hot path.

## Shape 2 — a mutual trampoline compared an owned temporary to a borrowed invariant

```aver
fn kept(item: List<Int>, rest: List<List<Int>>, target: List<Int>, acc: List<List<Int>>) -> …
    match List.reverse(item) == target
```

`error[E0308]: expected AverIntList, found &AverIntList`.

Worth recording, because it is the more interesting half: this is the **same class as #1037, #1060 and #1065, and it bypassed their fix entirely**. Mutual TCO does reach `mir_emit_equality`, but a trampoline's hoisted invariants live only in `rc_wrapped`, while the alignment logic added by those tickets consulted `borrowed_params` alone. Their tests covered ordinary borrowed parameters, never a mutual trampoline, so the gap was invisible.

Mutual trampoline contexts now classify `rc_wrapped` invariants as borrowed references for equality emission. The existing alignment then emits `&(item.reverse()) == target` — no allocation, no clone. Self-TCO's `Arc<T>` representation is untouched.

## Tests

Two regression tests in `tests/rust_codegen_differential.rs`, each red before its fix and green after. Both go further than asserting emitted text: they build the generated project and run its verify cases. The mutual test requires a `__MutualTco` trampoline in the output, because a self-recursive case never reaches that path — which is exactly how the earlier tickets missed this.

`cargo fmt --all -- --check`, `cargo clippy` and `cargo test --test rust_codegen_differential` (56 passed) all pass. Both reproducers from the issue compile and build end to end; verified independently of the implementer's report.

### DIFF
--- a/src/codegen/rust/from_mir.rs
+++ b/src/codegen/rust/from_mir.rs
@@ -108,6 +108,12 @@ pub struct MirEmitCtx<'a> {
     pub local_types: &'a HashMap<String, Type>,
     /// Params passed as `Rc<T>` (self-TCO) / `&T` (mutual-TCO).
     pub rc_wrapped: &'a HashSet<String>,
+    /// Whether `rc_wrapped` params are represented as `&T`. This is true
+    /// only while emitting mutual-TCO trampoline arms; self-TCO wraps the
+    /// same policy class in `Arc<T>` instead. Equality uses this distinction
+    /// to align an owned operand with a borrowed invariant without changing
+    /// clone policy for either TCO shape.
+    pub rc_wrapped_are_borrowed_refs: bool,
     /// Params emitted as `&T` (borrow-by-default for non-Copy,
     /// non-Str params).
     pub borrowed_params: &'a HashSet<String>,
@@ -173,6 +179,7 @@ impl<'a> MirEmitCtx<'a> {
             codegen: None,
             local_types: EMPTY_TYPES.get_or_init(HashMap::new),
             rc_wrapped: EMPTY_SET.get_or_init(HashSet::new),
+            rc_wrapped_are_borrowed_refs: false,
             borrowed_params: EMPTY_SET.get_or_init(HashSet::new),
             loop_carried_params: EMPTY_SET.get_or_init(HashSet::new),
             owned_params: EMPTY_SET.get_or_init(HashSet::new),
@@ -220,6 +227,7 @@ impl<'a> MirEmitCtx<'a> {
             codegen: Some(ctx),
             local_types: &policy.local_types,
             rc_wrapped: &policy.rc_wrapped,
+            rc_wrapped_are_borrowed_refs: false,
             borrowed_params: &policy.borrowed_params,
             loop_carried_params: empty_string_set(),
             owned_params: &policy.owned_params,
@@ -242,6 +250,7 @@ impl<'a> MirEmitCtx<'a> {
             codegen: Some(ctx),
             local_types: &policy.local_types,
             rc_wrapped: &policy.rc_wrapped,
+            rc_wrapped_are_borrowed_refs: false,
             borrowed_params: &policy.borrowed_params,
             loop_carried_params: empty_string_set(),
             owned_params: &policy.owned_params,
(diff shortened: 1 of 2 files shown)


n1bor:
So quick to fix! Thanks!

jasisz:
Quick to break, quick to fix! To be honest I had some gut-feeling this may happen after some of the last PRs aligning MIR for Rust codegen. I should have checked on btc-listener myself ;)
I am slowly preparing for 0.,29 release, which will unfortunately include one more breaking change - I will mention you in there for sure.

PR #1109, opened 2026-08-24
Align stdlib naming and make partial APIs honest

### LINKED ISSUE #1100: Bytes has no way to and from String: a UTF-8 encode and decode are missing
## What we are building

n1bor/btc-listener keeps a Bitcoin chain's Index in a key-value database behind a capability contract (`Infra.Kv`). The keys and values are built as `String` in pure domain code — `o:<txid>:<index>`, hex-encoded scripts — and the database is bytes. Today the contract says `String` at the boundary and the Rust provider converts with `String::from_utf8`. We want the contract to say `Bytes`, so the provider stops deciding what is text, and so the keyspace can later become binary without a second contract change.

## What is missing

`Bytes` can be made from a `List<Int>` or a hex `String`, and turned back into either:

```
Bytes.fromList : List<Int> -> Result<Bytes, String>
Bytes.toList   : Bytes -> List<Int>
Bytes.fromHex  : String -> Result<Bytes, String>
Bytes.toHex    : Bytes -> String
```

There is no way to get the UTF-8 encoding of a `String` as `Bytes`, or to decode `Bytes` that are UTF-8 back into a `String`. The runtime does exactly this inside `Tcp.send` / `Tcp.readLine` (text-only, UTF-8) and `Disk.readText`; the language does not expose it.

The only route today is character by character:

```aver
Bytes.fromList(codesOf(String.chars(text), []))      -- Char.toCode per character
asciiTextInto(Bytes.toList(bytes), "")              -- acc + Char.fromCode per byte
```

which allocates a `List<String>`, a `List<Int>` and then the `Bytes` for every key, concatenates a `String` once per byte on the way back, and is only correct for ASCII. On a write path of two hundred million entries it is the difference between the database being the bottleneck and the conversion being the bottleneck — so the contract stays `String` and the provider keeps deciding.

## The ask

```
Bytes.fromString : String -> Bytes                    -- the UTF-8 encoding; total
String.fromBytes : Bytes -> Result<String, String>    -- decode, Err on invalid UTF-8
```

Total one way because every `String` has a UTF-8 encoding; a `Result` the other because not every byte sequence is one, and the caller wants the reason. `String.byteLength` already counts UTF-8 bytes, so the runtime has the encoding in hand.

## What we are not asking for

Not asking for `Bytes` to become a `String` subtype, for implicit conversion at a capability boundary, or for any other encoding than UTF-8.

### LINKED ISSUE #1102: Names for conversions, properties and renderings follow no rule, and one namespace has no type
`to…` currently covers three unrelated operations, and one namespace names a type that does not exist. The result is that names carry no reliable information about what an operation does.

In `stdlib/bytes.av`, side by side:

- `fromList` walks the whole list, validates `0..=255`, introduces an invariant, returns `Result`
- `toList` is `bytes.values` — hands back a field, drops the invariant, cannot fail
- `toHex` builds a new representation character by character

`toList` and `toHex` share a prefix and share nothing else. `fromList`/`toList` read as a round trip and are not one. The same collision appears in `examples/`: `Json.toString` renders, `Email.toString` returns `e.address`.

## The rule

**A conversion is named on the type that comes into existence, when the author of the API controls that namespace. When it does not, it is named on the source.**

The earlier formulation said "when the target is builtin and closed", and `Option.toResult` is what disproved it: `Result` is builtin, so that wording forbade `Result.fromOption`, which is plainly the better name. The compiler controls every builtin namespace — that is why `Int.fromFloat` and `Vector.fromList` are already correct — while a user module cannot add to `String`, which is why `Address.toString` in a user project is right rather than a deviation.

Four forms follow:

1. `Target.fromX` — construction, named on what comes into existence
2. `Source.toFormat` — rendering into a *named* format, where a format is one representation among several: `Bytes.toHex`, and by the same reasoning `String.toUtf8` (see #1100)
3. bare property noun — the value handing back what it holds: `Bytes.octets`, alongside the existing `Map.keys`, `String.chars`. Says what is inside; claims nothing about cost
4. domain verb or noun — derivation, a genuinely new value: `Crypto.sha256`, `RedBlackTree.valuesInOrder`

A namespace is a semantic owner and need not name a type: `Bits`, `Console`, `Http` stay. `Char` is the single exception, and not because it names no type — because its two functions operate on `String`.

Not covered here: signatures that declare totality and raise at runtime (`Char.toCode` on `""` is one of a family). That is a correctness matter and gets its own issue.

## Renames

Counts are our whole tree / n1bor/btc-listener.

| today | becomes | uses |
|---|---|---|
| `Bytes.toList` | `Bytes.octets` | 10 / 161 |
| `Digest32.toBytes` | `Digest32.bytes` | 2 / 17 |
| `Natural.toInt` | `Natural.value` | 3 / 0 |
| `Positive.toInt` | `Positive.value` | 0 / 0 |
| `IntRange.toInt` | `IntRange.value` | 0 / 0 |
| `NonNegFloat.toFloat` | `NonNegFloat.value` | 0 / 0 |
| `Email.toString` | `Email.address` | 0 / 0 |
| `Option.toResult` | `Result.fromOption` | 51 / 0 |
| `Char.fromCode` | `String.fromCodePoint` | 61 / 1 |
| `Char.toCode` | `String.firstCodePoint` | 47 / 13 |
| `RedBlackTree.toSorted` | `RedBlackTree.valuesInOrder` | 39 / 0 |

`Char.toCode` is a rename here and a signature repair in the separate issue; the two land together but are justified separately.

## Deliberately unchanged

`String.fromInt`, `String.fromFloat`, `String.fromBool` — the compiler controls `String`, so target-owned naming is what the rule prescribes. There is no `Int.toString`; one name per operation.

`String.toLower`, `String.toUpper` — transformations within a type, and `lower`/`upper` would be more ambiguous, not less.

`Bytes.fromList`, `Bytes.fromHex`, `Bytes.toHex`, `Digest32.fromBytes`, `Digest32.fromHex`, `Digest32.toHex`, `Json.fromString`, `Json.toString`, every capability operation, `Option.Some`, `Result.Ok`, all queries and arithmetic.

Pairs where both sides are builtin and neither is the obvious subject — `Float.fromInt`, `List.fromVector`/`Vector.fromList` — are left as they are. The rule has nothing to say about them and inventing an answer would be worse than admitting that.

## Scope

Normative for this repository's builtin surface and standard library. Not a lint, and not normative for user code — a user project's own conventions are its own business.

### LINKED ISSUE #1103: Builtins that declare a total signature and then fault, including one that does not terminate
Several builtins declare a total signature and then raise a runtime error for input the type system accepts. `aver check` passes, `aver verify` reports `case aborted`, and the caller has no value to match on — which contradicts the rule that a failure is a value.

This is not a naming matter (#1102 covers those), though one entry appears in both: `Char.toCode` is renamed there and repaired here.

## Confirmed by running it

**`Char.toCode : String -> Int`** (`src/types/checker/builtins.rs:541`, `src/types/char.rs:42`) raises on `""`. It also silently returns the first code point of a longer string, so `Char.toCode("abc") => 97` passes verify green.

```
fail[verify-runtime-error]: case aborted
  error: Runtime error [line 9]: Char.toCode: string is empty
```

A user has already had to work around it in prose. From a real project's `domain/json.av`:

> "Never asked about an empty string: every caller has taken this character off the front of a list, and Char.toCode of nothing faults."

A precondition the type cannot carry, documented by hand. Repair: `String.firstCodePoint : String -> Option<Int>`.

**`Vector.new : (Int, T) -> Vector<T>`** (`src/types/checker/builtins.rs:508`, `src/types/vector.rs:45`) raises for a negative size, or one beyond what `usize` can address. Repair: `Result<Vector<T>, String>`.

**`Bits.shiftLeft`** is worse than a wrong signature. It is declared `Result<Int, String>` (`src/types/checker/builtins.rs:354`), which is right, but a shift count beyond machine range does not return `Err` — it starts computing. `Bits.shiftLeft(1, 4294967296)` did not finish within two minutes, and `Int` is arbitrary-precision, so the work is bounded only by memory. That is a denial of service reachable from ordinary source, and the same shape we reported upstream to a prover last month. `Bits.shiftRight` and `Bits.low` share the argument path (`src/types/bits.rs:91`).

## Confirmed by reading the implementation

**`Env.set : (String, String) -> Unit`** (`src/types/checker/builtins.rs:220`) — the host call's error becomes a runtime error (`src/services/env.rs:78`).

**`Terminal.moveTo : (Int, Int) -> Unit`** (`:247`) — raises for coordinates outside `i64`, and the adapter's own failure is mapped straight to a runtime error (`src/services/terminal.rs:113`).

**`Terminal.size : () -> Terminal.Size`** (`:277`) — an I/O query with no failure in its type; the whole `Terminal` family goes through `f().map_err(RuntimeError::Error)` (`src/services/terminal.rs:93`).

## Reported but not yet verified here

`BranchPath.parse`, `BranchPath.child`, `Random.int`, `Time.sleep`, `HttpServer.listen`, `HttpServer.listenWith`. Same suspected shape — a declared-total signature over a partial operation — but I have not run them, and the list should be confirmed before any of it is repaired.

## Why this matters beyond tidiness

A signature is what a caller reasons against, and what we export to a prover. `String -> Int` states that every string has an answer. When the implementation disagrees, the claim we certify is about a function that does not exist. The effect classification already forces an author to declare what a function touches; the return type should be equally honest about whether it can decline.

The repairs are not uniform — some want `Option`, some `Result`, and `Bits` wants a bound before the work starts, not a wrapper around it. Each changes call sites, so this is a semantic change and not a sweep.

jasisz:
## Summary

- add linear UTF-8 conversion as `String.toUtf8` and `String.fromUtf8` across the VM, Rust, wasm-gc, and wasip2 paths
- apply the naming rule from #1102 (`Bytes.octets`, `Digest32.bytes`, `Result.fromOption`, `String.firstCodePoint`, `String.fromCodePoint`, and project-level property renames)
- return `Option` or `Result` from builtins that previously declared total signatures and then faulted, including Vector, BranchPath, Env, the complete Terminal family, HttpServer, Random, and Time
- bound only the Bits operations that can actually materialize a growing result; `shiftRight` remains unbounded for non-negative counts
- discharge impossible validation errors for statically safe literal calls, including Int division/modulo, Bits counts, Vector sizes, BranchPath values, Random bounds, Time durations, and literal refinement constructors
- keep discharged effect calls real: `Random.int(1, 6)` still invokes and records/replays the provider, and `Time.sleep(1)` still performs the sleep
- migrate the full tree and synchronize provider contracts, Oracle/replay, self-host output, proof backends, examples, playground sources, documentation, and the CHANGELOG migration table

## Semantics

```aver
roll = Random.int(1, 6)  // Int; the Random.int effect still runs
Time.sleep(1)            // Unit; the Time.sleep effect still runs
```

Dynamic or invalid inputs keep their `Result` type. Literal discharge uses a private fail-closed unwrap: if a provider returns `Err` after its arguments were proven valid, execution faults as a provider-contract violation. The compiler never uses `Result.withDefault` to invent a random value or successful sleep; the public helper remains available only for explicit user-chosen fallback semantics.

All `Terminal.*` host adapters expose the same fallible boundary. This removes the inconsistent shape where `moveTo` returned `Result` but an adjacent `print`, `flush`, or mode/color operation could fault outside the language value.

`Bits.shiftRight` rejects negative counts but has no materialization ceiling. The 16,777,216-bit bound remains on `shiftLeft`, and on `low` only where the result can grow. `AVER_NO_PACKED_SEQUENCES` is removed; the boxed wasm representation is exercised through an internal test configuration instead of a process-global environment switch.

UTF-8 conversion is O(n): packed Rust and wasm paths perform a linear copy/validation without a code-point-list detour.

The Terminal result-shape changes and `Time.sleep` recording change are intentionally not replay-compatible with recordings made under the old contracts.

## Validation

- `git diff --check`
- `cargo check --locked --features wasm,wasip2,terminal`
- `cargo clippy --workspace --all-targets --exclude aver-lang -- -D warnings`
- `cargo clippy -p aver-lang --lib --bin aver -- -D warnings`
- `cargo test --locked --lib --features wasm,wasip2,terminal` — 1477 passed
- `cargo test --locked --test proof_spec --features wasm,wasip2,terminal` — 306 passed
- generated Lean project for `oracle_independent_products.av` — build passed with 0 errors and 0 `sorry`
- all 38 certificate scenarios passed in the earlier full workspace gate
- targeted VM, provider, replay, Rust codegen, wasm-gc, wasip2, Lean, and Dafny regressions passed

Closes #1100
Closes #1102
Closes #1103


n1bor:
Looking forward to this merge.
Then I have zero pending issues with aver. (For now)..

jasisz:
Thanks! This is the breaking change I mentioned earlier. If you would like, we can open a migration PR against `n1bor/btc-listener` once this merges.

n1bor:
Am fine on you just merging. My CI etc.. is pinned to a specific Aver release.
So once you merge I will find a break in other development and move onto this release and fix all the issues.

jasisz:
Addressed the contract review in `068e2d72`:

- made the complete `Terminal.*` family expose adapter failure as `Result`, including Oracle/replay and hostile profiles
- replaced compiler-generated `Result.withDefault` discharge with a private fail-closed intrinsic; a provider `Err` after valid literal arguments now faults instead of fabricating a value
- removed the materialization limit from `Bits.shiftRight` and limited `Bits.low` only where the result can actually grow
- removed `AVER_NO_PACKED_SEQUENCES` in favor of an internal boxed-representation test hook, including multibyte UTF-8 coverage
- added the CHANGELOG migration table and documented the intentional replay incompatibility
- carried the fail-closed boundary through Lean and Dafny without adding axioms or `sorry`

Validation after the fixes: 1477 library tests and all 306 `proof_spec` tests pass; the additional Lean independent-products build also passes with 0 errors / 0 `sorry`.

jasisz:
Follow-up review addressed in 8593a1a:

- migrated the canonical .claude/commands/aver.md guide to the renamed APIs, current Result shapes, and current Oracle classifications;
- added String.toUtf8 / String.fromUtf8 explicitly under CHANGELOG Added;
- made the hidden boxed-sequence test hook reject every non-wasm-gc compile target instead of silently doing nothing;
- renamed the Lean UTF-8 helpers to stringToUtf8 / stringFromUtf8 and pinned those names in the proof export regression;
- removed the dead Char. purity prefixes;
- corrected the stale README Random.int Oracle example to keep the stub response as Result.

Validation:
- cargo fmt --all -- --check
- cargo clippy -p aver-lang --lib --bin aver -- -D warnings
- cargo clippy --workspace --all-targets --exclude aver-lang -- -D warnings
- cargo test --locked --lib --features wasm,wasip2,terminal (1478 passed)
- focused Lean/Dafny embedded Bytes proof export test passed

The changes that need an actual policy or backend mechanism are tracked separately:
- #1110 Vector.new materialization policy
- #1111 single-source Bits materialization bound
- #1112 wasm-gc provider-contract trap diagnostics
- #1113 checked-in self-host freshness gate

Replay schema remains intentionally unchanged: this PR already documents that old suppressed/unit recordings are not replay-compatible, and supporting old recordings is outside its scope.

PR #1214, opened 2026-08-29
Keep String.eq certificate fixture canonical

jasisz:
## Summary

- restore the String.eq certificate fixture to the canonical one-comparison shape covered by the byte-level wall
- keep all classification, runtime-contract, and generated-plan assertions active when Lean is unavailable; only skip the final kernel build without `lake`

## Why

#1210 correctly disabled cursor fabrication for `--certify`, but the same follow-up commit also added a redundant multi-character control arm to `tools/certkit/fixtures/stringeq.av`. The MIR emitter faithfully preserved that extra `String.eq` cascade, while the certificate wall deliberately supports the previously covered one-comparison template. The wall therefore failed closed and classified only `bump`.

The control arm is no longer needed: with `run_chars_fusion: false` on certified artifacts, the original one-literal fixture emits `String.eq` and directly tests the certification seam. It restores the previously covered 6,680-byte artifact (`952613228bd9857089cfb5b387940442e1f9edf1547d2623fa66932ffaf9dfb0`).

Moving the `lake` availability check after the manifest/plan assertions makes this regression visible in ordinary CI environments that do not install Lean.

## Validation

- `cargo test --features wasm,wasip2 --test cert_certify_spec certify_string_eq_host_contract_lake_builds_kernel_clean -- --exact --test-threads=1`
- the same test binary with `lake` absent from `PATH` (classification and plan checks pass; kernel build alone is skipped)
- `git diff --check`

### DIFF
--- a/tests/cert_certify_spec.rs
+++ b/tests/cert_certify_spec.rs
@@ -2404,11 +2404,6 @@ fn certify_verbatim_variant_dispatch_lake_builds_kernel_clean() {
 
 #[test]
 fn certify_string_eq_host_contract_lake_builds_kernel_clean() {
-    if Command::new("lake").arg("--version").output().is_err() {
-        eprintln!("skipping certify String.eq host-contract test: `lake` not available");
-        return;
-    }
-
     let repo_root = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
     let out_dir = temp_dir("certify-stringeq");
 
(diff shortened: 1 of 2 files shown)


jasisz:
Post-merge `cert_decode_spec` exposed that its ordinary-runtime F5 mutation had also reused the certificate fixture. The runtime-only control source is split out in #1216; the previously failing case passes locally.

Issue #1254, opened 2026-09-03
wasm-gc: emit a self-contained Wasmtime host pack with providers


jasisz:
## Problem

`aver run --wasm-gc` can assemble a Wasmtime execution environment with Aver's standard providers and the custom providers selected by `aver.toml`, but that environment is currently a development-time service of the `aver` executable.

A raw wasm-gc module is therefore not a deployable server artifact by itself. Bare `wasmtime` cannot satisfy Aver's `aver/*` host ABI or the contract-derived custom capability imports, while requiring the full compiler CLI on a production machine is unnecessary and makes deployment less explicit.

`btc-listener` is the concrete motivating program: its wasm is useful outside a JS host, but it needs standard capabilities plus custom `Domain.Primitives` and `Infra.Kv` providers.

## Proposed product shape

Add a self-contained Wasmtime host pack, provisionally:

```bash
aver compile main.av --target wasm-gc --pack wasmtime
```

producing something like:

```text
out/
  main.wasm
  aver-host
  manifest.json
```

`aver-host` should contain Wasmtime, the standard native providers, and the project-specific custom providers resolved at pack time. Running `./aver-host [args...]` must not require `aver`, Cargo, or the Aver source tree on the destination machine. A later single-file form may embed `main.wasm`, but the directory form keeps the exact wasm artifact inspectable and replaceable under an explicit manifest binding.

The implementation should extract and reuse the provider assembly already exercised by `aver run --wasm-gc`; it should not grow a second host/runtime path.

## Required properties

- The manifest binds the exact wasm bytes, target ABI, provider identities, and capability contract/model hashes.
- The host rejects a module/provider mismatch before calling the guest.
- Standard and custom providers have the same behavior as `aver run --wasm-gc`, including policy and record/replay where enabled.
- The host is built for an explicit native target triple; the wasm remains the same portable guest artifact.
- `btc-listener` can run through the pack with no `aver` executable in `PATH`.
- CI exercises at least one standard-provider program and one custom-provider program from the emitted pack.

## Boundaries

This is not a replacement for `--target wasip2`: wasip2 is the standards-based component deployment path, while a self-contained Wasmtime host is needed when the program uses Aver-specific/custom capability imports or operations outside the available WASI worlds.

It is also not the generated-Rust backend. The delivered program remains the exact wasm-gc module and runs through the same Wasmtime/provider boundary used during development.

jasisz:
A note on ordering before anyone picks this up.

[...]

PR #1257, opened 2026-09-03
perf(wasm-gc): improve packed-key maps and Wasmtime GC throughput

jasisz:
## Summary

- hash proof-packed sequence Map keys with FNV-1a instead of DJB2, avoiding low-bit clustering in the power-of-two open-addressing table
- give every embedded Wasmtime path one shared wasm-gc engine configuration
- select the copying collector explicitly and start its heap at 16 MiB to avoid repeated growth/collection while allocation-heavy programs discover their working size
- use that configuration from `run --wasm-gc`, `run --wasip2`, wasm-gc verification, and the Wasmtime benchmark runner

The initial GC heap is virtual space that Wasmtime commits lazily; it does not eagerly consume 16 MiB of physical memory per Store. Keeping the engine configuration in one constructor also gives the future host-pack work in #1254 one configuration to extract/reuse.

### Packed `Map<Bytes, Int>` keys

btc-listener-shaped workload: 200,000 resident 37-byte keys and 2,000,000 probes, Node 26/V8, `--optimize speed`, seven-run medians:

| operation | DJB2 | FNV-1a |
|---|---:|---:|
| build | 139.5 ms | 64.8 ms |
| hits | 1052.0 ms | 805.7 ms |
| misses | 1555.2 ms | 515.7 ms |

### Full btc-listener validation

The application-shaped workload decodes and validates a real 1,911-byte Bitcoin Core transaction with twelve inputs and twelve outputs, executes six P2PKH and six P2WPKH scripts with actual sighash/signature checks, then applies the transaction against a twelve-entry UTXO view. Each sample warms five iterations and times fifty; all backends returned checksum `56485`.

| backend | median for 50 iterations |
|---|---:|
| generated Rust release | 112 ms |
| Wasmtime before heap tuning | 389 ms |
| Wasmtime with 16 MiB initial heap | 162 ms |
| V8 / Node 26 | 747 ms |

The V8 deployment uses `@noble/curves`, while Wasmtime and generated Rust use the btc-listener `libsecp256k1` provider, so the cross-backend table describes the real deployment stacks rather than an isolated JIT comparison. The 389 -> 162 ms Wasmtime comparison uses the same module and provider and isolates the engine configuration change.

## Verification

- `cargo fmt --all`
- `cargo check --features wasm`
- `cargo check --features wasip2`
- `cargo check --features wasm,wasip2`
- `cargo test --features wasm --lib codegen::wasm_gc::body::hash_helpers::tests` (3 passed)
- `cargo test --features wasm --test wasm_gc_spec` (90 passed)
- `cargo test --features wasm --test wasm_gc_verify_bool_case_regression --test wasm_gc_record_replay_spec` (17 passed)
- `cargo test --features wasip2 --test wasip2_poc` (3 passed)


jasisz:
### Follow-up: Binaryen `-O3` on the real btc-listener workload

I wired `aver run --wasm-gc --optimize speed` through the real cached custom-provider host in stacked PR #1259 and repeated this PR's full validation benchmark (1,911-byte / 12-input Bitcoin Core transaction, complete P2PKH + P2WPKH Script/ECDSA checks, then UTXO connect).

Seven alternating runs, 5 warmups + 50 timed iterations each; checksum was `56485` in every run:

| embedded Wasmtime | median workload | delta | whole-command wall |
|---|---:|---:|---:|
| emitter output | 164 ms | baseline | 2.208 s |
| `wasm-opt -O3` | 146 ms | **-11.0% / 1.12x** | 2.659 s |

The module shrank from 187,781 to 153,050 bytes (-18.5%). The startup result matters: invoking Binaryen on every ad-hoc `aver run` adds about 0.45 s, so O3 should remain opt-in there. For the proposed self-contained Wasmtime pack/host (#1254), it is a clear win: optimize once during pack, then ship/cache those bytes and retain the 11% steady-state improvement without paying Binaryen on every launch.

Issue #1376, opened 2026-09-15
Prove source request-trace equivalence for yield lowering


jasisz:
Follow-up to #1329, after the generated finite-history invariants in #1375.

The missing theorem compares the original direct-style function with its lowered process protocol under the same oracle. For every finite observable prefix they must produce the same ordered `(operation, arguments, answer)` events, allowing internal yields, and the same result when both reach `Done`. Equal final results alone are insufficient.

Original implementation prerequisites (source retention and dynamic trace observers are now implemented by #1377/#1378):

- `yield_lowering::lower` consumes the stamped source functions and retains protocols/signatures, but not source bodies for a second proof model. Preserve the original function and its module ownership before replacing it; do not reconstruct the source side from the generated machine.
- `types/checker/effect_lifting.rs` primarily assigns static call-site counters. Match arms advance the same compile-time counter, and ordinary helper calls do not return an updated occurrence counter. The special `Process.stopRequested` path threads a base, but is not a general process trace model. The source observer and protocol driver need dynamic executed-call positions, including repeated requests and in-place effects.
- `main/commands.rs` deliberately refuses proof export of process cases with `process_verification`; keep that refusal until the corresponding Oracle model is implemented. The VM driver in `yield_lowering/verify.rs` already executes the actual generated protocol and uses runtime stub counters; it is not an independent source specification.

Implementation / acceptance:

- [x] Retain independently readable source definitions and precise origin mapping through module loading, including private and imported yielding helpers.
- [x] Define concrete typed request/answer observations and the source and protocol transition semantics in ordinary generated Aver; no public user generics or handwritten example-only Lean companion.
- [ ] Thread dynamic Oracle state through branches, helpers and resumes. Internal yields consume no answer. State the in-place-effect and runtime conformance assumptions explicitly.
- [ ] Prove local source-to-protocol correspondence and compose it over arbitrary finite derivations/prefixes, with explicit helper-theorem dependencies. Different internal step counts must not be conflated with a common fuel number.
- [ ] Check the theorem using normal `aver proof --check`, with universal manifests and the existing axiom whitelist. Unsupported shapes must be explicit; sampled schedules cannot earn universal credit.
- [ ] Cover repeated calls of one kind, mixed kinds, branches, self yields, local/imported nested helpers, early completion, and in-place effects.
- [ ] Negative controls must reject dropped, duplicated, reordered and argument/answer-swapped requests even when the final return value remains unchanged; also catch Oracle counter resets across resumes.

[...]

PR #1394, opened 2026-09-17
feat(proof): observe imported in-place protocol effects

jasisz:
Imported yielding helpers with in-place capability calls cannot currently participate in source/protocol trace checking: their effectful Start/Answer functions lack an owning-module observation interface.

Export pure observations of the actual generated segment bodies. Typed import adapters preserve the caller’s event prefix, operation position, consumed count and original remaining tokens. Retained-source observation stays independent of protocol instrumentation.

Every module with a requested trace also checks an interface for each observed segment where the segment is defined: the observation leaves the tape at the suffix its own cursor reports, its incoming event history is only a prefix, and one protocol step is that observation followed by the generic continuation (the start segment has no step). The owner publishes a sample of each protocol state an importer has to quantify over. The statements are made about wrappers rather than about the observations themselves: a function that owns a law keeps its match equations, and every proof that opens it would then have to normalize a dependent matcher.

A caller reads this interface instead of an imported observer’s body. Its mapping law cites the cursor, prefix and step of every observation it drives through. Its source correspondence crosses one call site at a time, naming the segment observation and the drive result at that site, folding the protocol side into the site’s splice wrapper by the splice law and opening the wrapper. Its cursor contract cites the lifted cursor laws and proceeds by functional induction on its own protocol observer. Nothing owned by another module is unfolded. See `docs/yield-request-traces.md`.

Draft implementation for #1376. The fixture covers private recursion, nested and repeated imports, clock observations before and after requests, tail entry, caller-only input kinds and early errors. It requires 92 universal laws: 41 correspondence, mapping, splice and cursor laws, and 51 for the segment interface, plus negative controls for dropped/reordered events and reset positions.

Validation: CI (including WASM), Certification and Proof pass at ed460f7a. The other request-trace fixtures are unchanged. The temporary remote diagnostic workflow is removed in the last commit (12eb3fad). Main through #1401 (9c73e1ee) has been merged into the branch.

https://claude.ai/code/session_017KRynUsPVmwFP9ZcdwVyG5


jasisz:
Back to draft - I still need to work on this one.

